PR #18, opened 2018-03-19
fix: daemon type in tests

dryajov:


test/exchange-files.js
```diff
@@ -57,9 +59,9 @@ describe('exchange files', () => {
     this.timeout(50 * 1000)
 
     parallel([
-      (cb) => df.spawn({ initOptions: { bits: 1024 } }, cb),
-      (cb) => df.spawn({ type: 'js', initOptions: { bits: 512 } }, cb),
-      (cb) => df.spawn({ type: 'js', initOptions: { bits: 512 } }, cb)
+      (cb) => goDf.spawn({ initOptions: { bits: 1024 } }, cb),
+      (cb) => jsDf.spawn({ type: 'js', initOptions: { bits: 512 } }, cb),
+      (cb) => jsDf.spawn({ type: 'js', initOptions: { bits: 512 } }, cb)
```
diasdavid: Why does it needs `type` two times?

JonKrone:
Could we at some point remove the need to `create` altogether? `create`ing selects the type of daemon factory (client, in-proc, or go/js) but we can infer that based on `type` and environment just as we do now. Is there real utility from separating `create` from `spawn`ing the daemon?

dryajov:
The idea behind `.create` was to allow creating a concrete factory to spawn the different types of nodes. Although we can change it, there is little benefit in doing so now.

dryajov:
Ah, I see 😄 

The difference between the current `df.create({type: 'js', remote: true}).spawn()` or a possible `df.create().spawn({type: 'js', remote: true})` is minimal IMHO. It seemed cleaner to implement considering we're creating different factory types based either on parameters or environment.

JonKrone:
Yeah haha that is minimal difference but I guess I wasn't clear: I was suggesting `ipfsd.spawn({ ?type, ?remote, ... })`. So, `ipfsd.spawn` would basically do the same inference as `df.create` and then just additionally spawn the daemon.

With the current factory setup, we do get access to `tmpDir` and `version` before `spawn`ing but `version` spawns a daemon anyways and `tmpDir` could instead be a separately exported util, making the factories minimally useful.

PR #20, opened 2018-05-18
New ipfs

diasdavid:


diasdavid:
The reason why PubSub tests fail sometimes is because they haven't been updated to the latest API which now is inconsistent until js-ipfs gets released.

@alanshaw merging here. As part of the 0.29 release, let's create a PR here that updates the pubsub calls.

alanshaw:
@diasdavid I just circled back to open a PR here as requested but as far as I can tell, these tests won't be affected by the changes to the pubsub API.

`pubsub.subscribe` does not use `options` (it would have to be passing `options` to be affected by the argument ordering change) and `pubsub.unsubscribe` is never called.

Hooray?

diasdavid:
@alanshaw you are correct! Thank you for triple checking :)

PR #22, opened 2018-05-29
test: add test for mfs leaf node types

achingbrain:
Will not pass until MFS has been merged into js-ipfs.


achingbrain:
Closing in favour of #25 as I've added a bunch more tests.

PR #24, opened 2018-06-26
WIP: Private network tests

jacobheun:
This depends on https://github.com/libp2p/js-libp2p-pnet/pull/1 and its subsequent release into libp2p.

I created this to provide visibility into the existing interop tests while those all get merged and released.

Currently the only running tests are for js to js and js to go on the same private network. There is currently an issue in pull-reader with a private node attempts to talk with an invalid node (public or with a different key) that causes the initial connection to hang until it times out. I've added comments above the skips pointing to the issue. Rather than leave in tests that timeout, I've skipped them until we're able to error the connection out timely.


daviddias:
@jacobheun can you take these to the finish line in the foreseeable future?

jacobheun:
@daviddias yes! I updated the blocking PR in ipfs-repo if you can take a look https://github.com/ipfs/js-ipfs-repo/pull/172. There were concerns there about not using the keychain, but the spec isn't written to support using the keychain.

PR #26, opened 2018-07-07
test: add ipns tests

vasco-santos:
In the context of the `js-ipfs` [PR](https://github.com/ipfs/js-ipfs/pull/1400) for the `IPNS working locally` implementation, I made this PR.

PR's needed:

- [x] [js-ipfs-repo#173](https://github.com/ipfs/js-ipfs-repo/pull/173)
- [x] [js-ipns#4](https://github.com/ipfs/js-ipns/pull/4)
- [x] [js-ipfs#1400](https://github.com/ipfs/js-ipfs/pull/1400)
- [x] Release [`js-ipfs@0.32.0`](https://github.com/ipfs/js-ipfs/issues/1497) with `IPNS` and replace the package.json dependency (actually pointing to master)
- [x] [js-ipfs#157](https://github.com/ipfs/js-ipfs/pull/1570)


test/ipns.js
```diff
@@ -0,0 +1,328 @@
...
+          expect(res).to.equal(ipfsRef)
+          cb()
+        })
+      },
+      (cb) => jsDaemon.stop(cb)
+      // (cb) => setTimeout(cb, 10500)
+    ], done)
+  })
+
+  it('should publish an ipns record to a go daemon and resolve it using the same go daemon', function (done) {
```
vasco-santos: I am having strange problems with this test. I need to have two go daemons running, in order to be able to publish an IPNS record. Otherwise, I receive `Error: failed to find any peer in table`

test/ipns.js
```diff
@@ -0,0 +1,328 @@
...
+          expect(err).to.not.exist()
+          expect(res).to.equal(ipfsRef)
+          cb()
+        })
+      },
+      (cb) => goDaemon1.stop(cb),
+    ], done) */
+  })
+
+  it('should publish an ipns record to a js daemon and resolve it using a go daemon through the reuse of the same repo', function (done) {
```
vasco-santos: There is a problem with the repo in this test. I have also added a test for simulating this in the repo tests.

Basically, If we try to reuse to same repo previously used in JS (after running init) using a non-disposable daemon, the start fails with `Error: required Datastore.Spec entry missing from config file`. I tried to add the datastore config to the spawn method (in the repo test, but I am not sure if it is correct) but I couldn’t figure out the problem here.

test/ipns.js
```diff
@@ -0,0 +1,265 @@
...
+          console.log('res', res)
+          cb()
+        })
+      },
+      (cb) => goDaemon1.stop(cb),
+      (cb) => goDaemon2.stop(cb)
+    ], done)
+  })
+
+  it('should publish an ipns record to a go daemon and resolve it using a js daemon through the reuse of the same repo', function (done) {
```
vasco-santos: In this test, I can use the same repo (id is equal for both), but cannot resolve the record. I am not sure if it is a problem of the JS implementation, or the datastore. I would like to debug the GO part, but is is running in an EXEC file. Any suggestion for debugging in the go side?

I will analyze this deeper, in order to understand if this a problem with the IPNS implementation, or with the repo.

test/ipns.js
```diff
@@ -0,0 +1,252 @@
...
+const os = require('os')
+const path = require('path')
+const hat = require('hat')
+
+const DaemonFactory = require('ipfsd-ctl')
+const goDf = DaemonFactory.create()
+const jsDf = DaemonFactory.create({ type: 'js' })
+
+describe.only('ipns', () => {
+  it('should publish an ipns record to a js daemon and resolve it using the same js daemon', function (done) {
```
Stebalien: It would be nice to deduplicate this logic with the go test, if possible.

test/ipns.js
```diff
@@ -0,0 +1,252 @@
...
+          (cb) => goDaemon1.stop(cb),
+          (cb) => goDaemon2.stop(cb)
+          // TODO cleanup repo
+          // TODO check need for 2 daemons
+        ], done)
+      })
+    })
+  })
+
+  it('should publish an ipns record to a js daemon and resolve it using a go daemon through the reuse of the same repo', function (done) {
```
Stebalien: Again, this should probably be generic (so we can do a js resolution through a go node and vice-versa.

Stebalien:
> I am having strange problems with this test. I need to have two go daemons running, in order to be able to publish an IPNS record. Otherwise, I receive Error: failed to find any peer in table.

You'll need to start the daemon in offline mode for that to work (pass `--offline`).

Stebalien:
> There is a problem with the repo in this test. I have also added a test for simulating this in the repo tests.
> Basically, If we try to reuse to same repo previously used in JS (after running init) using a non-disposable daemon, the start fails with Error: required Datastore.Spec entry missing from config file. I tried to add the datastore config to the spawn method (in the repo test, but I am not sure if it is correct) but I couldn’t figure out the problem here.

The repos are currently not compatible. Go supports complex datastore configs and, as far as I know, javascript doesn't (yet).

Stebalien:
> In this test, I can use the same repo (id is equal for both), but cannot resolve the record. I am not sure if it is a problem of the JS implementation, or the datastore. I would like to debug the GO part, but is is running in an EXEC file. Any suggestion for debugging in the go side?

How are you storing the IPNS record currently? go-ipfs does will store the IPNS record in a libp2p "record" (https://github.com/libp2p/go-libp2p-record/blob/master/pb/record.proto) under `"/" + base32("/ipns/" + binaryPeerID)`.

vasco-santos:
@Stebalien 

> The repos are currently not compatible. Go supports complex datastore configs and, as far as I know, javascript doesn't (yet).

Yes, I already addressed that problem in [ipfs/interop#29](https://github.com/ipfs/interop/pull/29) and I am waiting for feedback.

> How are you storing the IPNS record currently? go-ipfs does will store the IPNS record in a libp2p "record" (https://github.com/libp2p/go-libp2p-record/blob/master/pb/record.proto) under `"/" + base32("/ipns/" + binaryPeerID)`.

Regarding this, I tried to reach you yesterday for discussing this.

I have been debugging the go implementation and I think that I understood the problem. Basically, as go-ipfs have a local + dht implementation, after publishing a record, you will have 3 distinct entries in your datastore. For instance, consider the following ones:

```
/F5UXA3TTF4JCA4L3WCGADRZNGPJYSNG5OBLYSSDFQKRNQZHCIMH24IZRO7MOH7K6
/F5YGWLYSEBYXXMEMAHDS2M6TRE2N24CXRFEGLAVC3BSOEQYPVYRTC56Y4P6V4
/ipns/CIQHC65QRQA4OLJT2OETJXLQK6EUQZMCULMGJYSDB6XCGMLX3DR72XQ
```

As I am doing the IPNS local implementation first, I only have the last one on js repo. 

Accordingly, when I run the following command in go:

`Ipfs name resolve —local`

I expect that it only cares about the last entry and it does not happen in fact. At first, it tries to get the PublicKey here (https://github.com/ipfs/go-ipfs/blob/master/namesys/routing.go#L76), which fails. I commented that block, since it only checks if the public key exists, and proceeded, in order to understand if it was the only problem. However, I got another problem in here (https://github.com/ipfs/go-ipfs/blob/master/namesys/routing.go#L86).

Basically, both problems try to get from the datastore what was stored in the routing context: https://github.com/ipfs/go-ipfs/blob/master/namesys/publisher.go#L225 

As it was the routing context, I didn’t implement that function, but it will be quick to add it if it is supposed. This way, I only want to ask you if this is in fact the correct approach, because if we run locally, we will immediately have the public key and the local record. Am I right, or missing something?

Moreover, I found another interop problem that I will work on today. I only noticed during this debug that you also publish an IPNS record during the `init` command, and I also have to add that part to `JS` pull request. Any more hidden usages of IPNS?

Stebalien:
> As I am doing the IPNS local implementation first, I only have the last one on js repo.

That last one isn't actually a part of the "routing" system, we put raw IPNS records there for republishing (both to remember the last-used sequence number and to remember which records we need to periodically republish). 

The first record in that list is the IPNS record as stored in the routing system. That's the one we care about here.

The second record is the public key record. Now that we (a) embed the public key in the IPNS record and (b) don't sign the outer "routing" record, that's less important.

Stebalien:
> I commented that block, since it only checks if the public key exists, and proceeded, in order to understand if it was the only problem.

See the comment above that line. We need the public key to validate the record. Note: once everyone starts just embedding the public key in their IPNS record, we won't need that. Really, we might just want to *not* fail if that lookup fails and try getting the IPNS record anyways.

Stebalien:
> I only noticed during this debug that you also publish an IPNS record during the init command, and I also have to add that part to JS pull request.

We publish the empty directory but that's not really part of the IPNS spec. That's just something that go-ipfs does. We probably do this because we support *mounting* user's IPNS entry using fuse (under `/ipns/local`). I wouldn't say that's a strictly necessary feature.

> Any more hidden usages of IPNS?

You'll have to make sure to republish IPNS entries once every ~4hr.

vasco-santos:
PR was refactored, and the interop was also achieved!! 🙌

PR's needed:

- [ ] [js-ipfs-repo#173](https://github.com/ipfs/js-ipfs-repo/pull/173)
- [ ] [js-ipfs#1461](https://github.com/ipfs/js-ipfs/pull/1461) (should also update the `js-ipfs-repo` version)
- [ ] [js-ipns#4](https://github.com/ipfs/js-ipns/pull/4)
- [ ] [js-ipfs#1400](https://github.com/ipfs/js-ipfs/pull/1400)

package.json
```diff
@@ -51,7 +51,7 @@
     "form-data": "^2.3.2",
     "go-ipfs-dep": "~0.4.17",
     "hat": "0.0.3",
-    "ipfs": "~0.31.6",
+    "ipfs": "github:ipfs/js-ipfs",
```
diasdavid: This has been released, you can :shipit:  :D

vasco-santos: Yeah! Just found a problem blocking the interop to work 😞 

[js-ipfs#1570](https://github.com/ipfs/js-ipfs/pull/1570)

alanshaw: @vasco-santos https://github.com/ipfs/js-ipfs/pull/1570 now merged and released, can this now be updated so we can merge?

test/ipns.js
```diff
@@ -0,0 +1,130 @@
...
+    (cb) => {
+      resolverDaemon.api.name.resolve(nodeId, { local: true }, (err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.equal(ipfsRef)
+        cb()
+      })
+    },
+    (cb) => resolverDaemon.stop(cb),
+    (cb) => setTimeout(cb, 2000),
+    (cb) => resolverDaemon.cleanup(cb)
```
alanshaw: Are we missing cleanup for the other daemon?

test/ipns.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+    series([
+      (cb) => publisherDaemon.stop(cb),
+      (cb) => setTimeout(cb, 2000),
+      (cb) => resolverDaemon.start(cb)
+    ], callback)
+  }
+
+  series([
+    (cb) => publisherDaemon.init(cb),
```
alanshaw: Are we missing init for the other daemon?

test/ipns.js
```diff
@@ -0,0 +1,130 @@
...
+      expect(err).to.not.exist()
+      publishAndResolve(goDaemon, done)
+    })
+  })
+
+  it('should publish an ipns record to a js daemon and resolve it using a go daemon through the reuse of the same repo', function (done) {
+    this.timeout(120 * 1000)
+    const dir = path.join(os.tmpdir(), hat())
+
+    parallel([
```
alanshaw: This has to be `series` for now, `ipfsd-ctl` will occasionally error for parallel spawns because of races for ports.

test/ipns.js
```diff
@@ -0,0 +1,130 @@
...
+        cb()
+      })
+    },
+    (cb) => resolverDaemon.stop(cb),
+    (cb) => setTimeout(cb, 2000),
+    (cb) => resolverDaemon.cleanup(cb)
+  ], callback)
+}
+
+describe('ipns', () => {
```
alanshaw: This could be named better - these are IPNS repo interop tests, when I first started reviewing I had incorrectly assumed they were testing that IPNS records published on js could be resolved on go and vice versa _over the network_.

test/ipns.js
```diff
@@ -0,0 +1,130 @@
...
+
+      publishAndResolve(daemons[0], daemons[1], done)
+    })
+  })
+
+  it('should publish an ipns record to a go daemon and resolve it using a js daemon through the reuse of the same repo', function (done) {
+    this.timeout(160 * 1000)
+    const dir = path.join(os.tmpdir(), hat())
+
+    parallel([
```
alanshaw: Also `series` :(

test/ipns.js
```diff
@@ -0,0 +1,130 @@
...
+  if (typeof resolverDaemon === 'function') {
+    callback = resolverDaemon
+    resolverDaemon = publisherDaemon
+    sameDaemon = true
+  }
+
+  const stopPublisherAndStartResolverDaemon = (callback) => {
+    if (sameDaemon) {
+      return callback()
+    }
```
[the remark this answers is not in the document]
vasco-santos: I agree!

test/ipns.js
vasco-santos: The cleanup is for the repo. 

I think that you did not understand the flow here, considering this comment and the next one. The flow is the following:

1) Init X daemon (with repo Z)
2) Start X daemon
3) IPNS Publish
4) Stop X daemon
5) Start Y daemon (with repo Z)
6) IPNS Resolve
7) Stop Y daemon
8) Clean repo Z

X and Y may be replaced for JS or GO.

This way, both implementations will use the same repo to publish and resolve the IPNS record, since we do not have routing yet.

test/ipns.js
vasco-santos: Flow explanation above.

test/ipns.js
vasco-santos: I changed it to: `ipns locally using the same repo across implementations`. What do you think?

test/ipns.js
alanshaw: Ah, yes I realised what was going on later and forgot to remove these comments. Please ignore!

PR #29, opened 2018-07-22
fix: repo interop

vasco-santos:
When the `js-ipfs-repo` adds the `datastore_spec` file, as well as the `datastore` property to the config, the repo's will be interoperable again 🚀 

Tested with both PR's needed:

- [x] [js-ipfs#1461](https://github.com/ipfs/js-ipfs/pull/1496)
- [x] [js-ipfs-repo#173](https://github.com/ipfs/js-ipfs-repo/pull/173)

### DIFF
--- a/test/repo.js
+++ b/test/repo.js
@@ -80,7 +80,7 @@ describe.skip('repo', () => {
   })
 
   it('read repo: js -> go', function (done) {
-    this.timeout(50 * 1000)
+    this.timeout(80 * 1000)
     const dir = path.join(os.tmpdir(), hat())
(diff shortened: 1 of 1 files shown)


vasco-santos:
After a productive call with @jacobheun we created an [endeavour issue](https://github.com/ipfs/js-ipfs/issues/1467) to tackle all the tasks regarding the repo interoperability. This issue aims to tackle the following set of tasks:

#### 4) [ipfs/interop](https://github.com/ipfs/interop)

- [ ] **1.**  Have current interop tests passing and not skipped

PR #33, opened 2018-08-29
test: re-enable pubsub tests

achingbrain:
Instead of using timeouts, explicitly connect nodes, subscribe to the topic then wait for the local node to appear in the remote node's peer list for a given topic before sending messages.


achingbrain:
Follows on from ipfs/interop#32

PR #36, opened 2018-09-17
test: add dht interop tests

vasco-santos:
This PR needs the following changes:

- [x] [js-libp2p-kad-dht/pull/45](https://github.com/libp2p/js-libp2p-kad-dht/pull/45)
- [x] [js-libp2p-kad-dht/pull/42](https://github.com/libp2p/js-libp2p-kad-dht/pull/42)
- [x] [js-libp2p-record/pull/8](https://github.com/libp2p/js-libp2p-record/pull/8)
- [x] [js-libp2p#251](https://github.com/libp2p/js-libp2p/pull/251)
- [x] [js-ipfsd-ctl#306](https://github.com/ipfs/js-ipfsd-ctl/pull/306)
- [x] [ipfs/js-ipfs-api#890](https://github.com/ipfs/js-ipfs-api/pull/890)
- [x] [js-ipfs#856](https://github.com/ipfs/js-ipfs/pull/856)


test/kad-dht.js
```diff
@@ -3,110 +3,516 @@
...
+
+const spawnJsDaemon = (bootstrap = [], callback) => {
+  jsDf.spawn({
+    initOptions: { bits: 512 },
+    config: getConfig(bootstrap),
+    args: ['--enable-dht-experiment']
+  }, callback)
+}
+
+const getNodeId = (node, callback) => {
```
jhiesey: Maybe call this getBootstrapAddr?

test/kad-dht.js
```diff
@@ -3,110 +3,516 @@
...
+    })
+
+    after(function (done) {
+      this.timeout(100 * 1000)
       series([
-        (cb) => goD1.stop(cb),
-        (cb) => goD2.stop(cb),
-        (cb) => goD3.stop(cb),
-        (cb) => jsD.stop(cb)
+        (cb) => node1.stop(cb),
```
jhiesey: I think this is fine, but it could be even cleaner if all the nodes were in an array.

test/kad-dht.js
vasco-santos: Better naming yes, I will change it!

vasco-santos:
Hey @jhiesey 

Thanks for your help.

> Looks good, just a few small comments. Do these tests pass at the moment?

The tests pass locally if you change the following [here](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/index.js#L116) (`node_modules`):

```js
start (callback) {
    this._running = true
    this.network.start((err) => {
      if (err) {
        return callback(err)
      }
      this.randomWalk.start(1, 100)

      callback()
    })
  }
```

Attention, this is not an appropriate change, as we should use the `randomWalk` defaults, but for that we have to change the way it is implemented with `setInterval`. I have started working on it yesterday, and hopefully today I will have a change to propose.

> Is it deliberate that there aren't tests that go over multiple hops (other than through the bootstrap node) anymore?

I would like to create the tests with multiple hops as they were before. However, I don't know how to create a test in that format. The previous implementation of that format connected all the peers in a sequence. That way, the test was running successfully with the DHT disabled, so I believe that it was not a valid test. If you have any network suggestion to achieve that, I would like to add it!

> I think the way it is is fine, but have you considered a templated version of this test that just lets you specify the types of node0 through node3 and then call a function to run the test? I'm not very familiar with mocha so I don't know if that's tricky or not.

I will have a refactor on this later, this is just a WIP for now, and before getting it ready for merging, I will try to reuse the most of the duplicated code.

jacobheun:
DHT interop tests look good. We need to get ci green before this is good. If there are flakey, unrelated tests we should get a PR to resolve those. Right now the job history for this PR has no successes https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Finterop/activity?branch=PR-36

vasco-santos:
@jacobheun in the last time that I ran the tests all were ok for the DHT tests, and the ones that failed were due timeouts:

https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Finterop/detail/PR-36/9/pipeline/361

This way, I wanted to have feedback about the PR before trying to have the all CI green.

test/kad-dht.js
```diff
@@ -1,112 +1,545 @@
...
+        const bootstrap = [bootstrapAddr]
+
+        parallel([
+          (cb) => spawnJsDaemon(bootstrap, cb),
+          (cb) => spawnJsDaemon(bootstrap, cb),
+          (cb) => spawnJsDaemon(bootstrap, cb)
+        ], (err, spawnedNodes) => {
+          expect(err).to.not.exist()
+          spawnedNodes.forEach((node) => nodes.push(node))
+          setTimeout(done, 25000) // Wait for peers to have each other on the DHT
```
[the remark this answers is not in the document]
alanshaw: ...you might want to steal https://github.com/ipfs/js-ipfs/blob/master/test/utils/wait-for.js

test/kad-dht.js
```diff
@@ -1,112 +1,545 @@
...
+          })
+        })
+      })
+
+      // spawn daemons
+      before(function (done) {
+        this.timeout(70 * 1000)
+        const bootstrap = [bootstrapAddr]
+
+        parallel([
```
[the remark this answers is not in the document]
alanshaw: \+ other occurances

test/kad-dht.js
```diff
@@ -1,112 +1,545 @@
...
+        ], (err, spawnedNodes) => {
+          expect(err).to.not.exist()
+          spawnedNodes.forEach((node) => nodes.push(node))
+          setTimeout(done, 25000) // Wait for peers to have each other on the DHT
+        })
+      })
+
+      after(function (done) {
+        this.timeout(80 * 1000)
+        parallel(nodes.map((node) => (cb) => node.stop(cb)), done)
```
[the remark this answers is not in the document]
alanshaw: \+ other occurances

test/kad-dht.js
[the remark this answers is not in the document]
alanshaw: \+ same thing for other occurances

test/kad-dht.js
```diff
@@ -1,112 +1,545 @@
...
+          nodes = spawnedNodes
+          done()
+        })
+      })
+
+      // make connections
+      before(function (done) {
+        this.timeout(60 * 1000)
+
+        parallel(nodes.map((node) => (cb) => node.api.id(cb)), (err, ids) => {
```
alanshaw: Suggestion: `async/each`

test/kad-dht.js
```diff
@@ -1,112 +1,545 @@
...
+          expect(err).to.not.exist()
+          parallel([
```
alanshaw: Made understanding the tests a whole lot easier!:
```
+-+       +-+
|0+-----> |1|
+++       +++
 ^         |
 |         |
 |         v
+++       +++
|3|       |2|
+-+       +-+
```

test/kad-dht.js
```diff
@@ -1,112 +1,545 @@
...
+    initOptions: { bits: 1024 },
+    config: getConfig(bootstrap)
+  }, callback)
+}
+
+const spawnJsDaemon = (bootstrap = [], callback) => {
+  jsDf.spawn({
+    initOptions: { bits: 512 },
+    config: getConfig(bootstrap),
+    args: ['--enable-dht-experiment']
```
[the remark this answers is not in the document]
vasco-santos: yes! This is outdated now

test/kad-dht.js
vasco-santos: In this case, I prefer this approach as using `each` would need to have an auxiliary array for keeping the ids for the second step.

test/kad-dht.js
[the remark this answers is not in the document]
vasco-santos: agree!

test/browser.js
```diff
@@ -3,7 +3,7 @@
 
 // require('./exchange-files')
 // require('./pubsub')
-require('./kad-dht')
+// require('./kad-dht')
```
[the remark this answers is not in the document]
vasco-santos: I have disabled them at the time because `daemon.api.id` was returning an empty array for addresses. This way, with no bootstrap peers we would take a long time in order to have the random walk discovery to find the remaining peers.

daviddias:
@vasco-santos mind rebasing master onto this PR and test it with js-ipfs master?

alanshaw:
@vasco-santos what's the status of this? Can we get CI green and merge?

vasco-santos:
I rebased this with master, but we are currently having failures that lead to `JavaScript heap out of memory` in the exchange files

jacobheun:
https://github.com/ipfs/interop/pull/59 should help mitigate the heap issue with exchange files. I think with the dht on, the data transfer across all the various nodes results in the heap issue. While the PR doesn't fix the underlying cause it should prevent it until we can resolve the issue.

test/browser.js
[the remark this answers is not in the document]
daviddias: Quick question: Are these tests using the Main Network to test (e.g. use public boostrappers)?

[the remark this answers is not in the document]
vasco-santos: No, in this case they are being connected only to each other

alanshaw:
Can has rebase and 💚 CI @vasco-santos?

vasco-santos:
Done @alanshaw 

However, as `ipns over dht`, this needs the DHT PR for enabling it on `js-ipfs`

daviddias:
@vasco-santos what is this blocked on?

vasco-santos:
It is blocked on having the DHT enabled. 

When we had the experimental flag for it, we could enable it through `ipfsd-ctl`. So, this PR is installing a branch with it enabled `js-ipfs#refactor/enable-dht`

vasco-santos:
Now we are able to also add the libp2p options through the `daemon.spawn` 🎉 

I will work on getting this updated with green CI

vasco-santos:
@alanshaw this should be ready to go. What do you think?

daviddias:
@vasco-santos mind rebasing master to confirm that CI is still green for clean merge?

vasco-santos:
ready to go @daviddias

test/browser.js
daviddias: This means it is not being tested

test/kad-dht.js
```diff
@@ -1,112 +1,692 @@
...
+}
+
+const getBootstrapAddr = (node, callback) => {
+  node.api.id((err, res) => {
+    expect(err).to.not.exist()
+    expect(res.id).to.exist()
+
+    callback(null, res.addresses[0])
+  })
+}
```
daviddias: This function is a tad misleading. All it does is extract the first listing address of a node. Perhaps rename to getNodeAddr?

test/kad-dht.js
```diff
@@ -1,112 +1,692 @@
...
+          expect(err).to.not.exist()
+          nodes.push(node)
+
+          getBootstrapAddr(nodes[0], (err, res) => {
+            expect(err).to.not.exist()
+            bootstrapAddr = res
+            done()
+          })
+        })
+      })
```
daviddias: It only tests the bootstrap with a JS Daemon, should be tested symmetrically (i.e. with go being the bootstrapper)

test/kad-dht.js
```diff
@@ -1,112 +1,692 @@
...
+
+      // spawn daemons
+      before(function (done) {
+        this.timeout(70 * 1000)
+        const peersToSpawn = 3
+        const bootstrap = bootstrapAddr ? [bootstrapAddr] : []
```
daviddias: In which case will bootstrapAddr not be available?

test/kad-dht.js
vasco-santos: Yeah, agree! I will change that!

test/kad-dht.js
vasco-santos: I am not sure if I understand your comment. But, I think I have those tests in the suite tests below

test/kad-dht.js
```diff
@@ -1,112 +1,692 @@
...
+
+        times(peersToSpawn, (_, next) => {
+          spawnJsDaemon(bootstrap, (err, node) => {
+            expect(err).to.not.exist()
+            nodes.push(node)
+            next(err, node)
```
daviddias: Why pass `err` here if it won't exist? Also, why pass node if you don't use the results out of the times call?

test/browser.js
vasco-santos: Yes, more info: https://github.com/ipfs/interop/pull/36#pullrequestreview-188608872

I can have a look after to that issue with `id()`, but I think that it should not be a blocker for this. I can create an issue to track it, as I think it is better to have these tests in now, than have nothing

vasco-santos: https://github.com/ipfs/interop/pull/36#discussion_r244681154

test/kad-dht.js
```diff
@@ -1,112 +1,692 @@
...
+          expect(err).to.not.exist()
+
+          const testDhtReady = (cb) => {
+            nodes[0].api.swarm.peers((err, peers) => {
+              if (err) {
+                return cb(err)
+              }
+              cb(null, peers.length === peersToSpawn)
+            })
+          }
```
daviddias: This function seems to only be testing if the first node is connected to all the other peers. It means that either:

- The name _testDhtReady_ is incorrect
- It is creating a full mesh which will then defeat a bit of the purpose of testing the DHT

test/kad-dht.js
daviddias: I see now below that the inverse happens

test/kad-dht.js
```diff
@@ -1,112 +1,692 @@
...
+        }, done)
+      })
+
+      it('should get from the network after being added', function (done) {
+        this.timeout(60 * 1000)
+
+        addFileAndCat(nodes[1], [nodes[2], nodes[3]], data, done)
       })
     })
   })
```
daviddias: @vasco-santos I believe the tests above would be better if the tests were described through which nodes play which role (who is the node providing the content, who is the node fetching the content, who is the node bootstrapping the network, who are the other nodes) and make it just a simple config rather than hard wiring things by hand in every test.

What do you think?

test/kad-dht.js
```diff
@@ -1,112 +1,692 @@
...
 
-  describe('a Go node in the land of JS', () => {})
-  describe('hybrid', () => {})
+  describe('kad-dht with connections per peer', () => {
```
daviddias: What does it mean "with connections per peer"?

test/kad-dht.js
```diff
@@ -1,112 +1,692 @@
...
+      })
+
+      it('three hops', function (done) {
+        this.timeout(85 * 1000)
+        const data = crypto.randomBytes(9001)
+
+        addFileAndCat(nodes[2], [nodes[3]], data, done)
+      })
+    })
+  })
```
daviddias: Again, this batch of tests could be more clean if instead of c&p the tests, they were only written once and then called multiple times with the different node configurations

achingbrain:
Superseded by #376

PR #39, opened 2018-10-27
test: add ipns pubsub tests

vasco-santos:
PR containing a set of tests aimimg to test the IPNS over Pubsub interoperability between `go-ipfs` and `js-ipfs`.

It needs the following:

- [x] `go-ipfs` released with [libp2p/go-libp2p-pubsub-router#17](https://github.com/libp2p/go-libp2p-pubsub-router/pull/17)
- [x] `js-ipfs` released with [ipfs/js-ipfs/pull/1559](https://github.com/ipfs/js-ipfs/pull/1559)

cc @bigs


vasco-santos:
Currently, we are facing a problem regarding the interop of the pubsub topic.

 `go-ipfs` sets the topic as `/ipns/BINARY_PEER_ID`. In Golang strings are a slice of bytes, and consequently, do not have a specific encoding. On the other side, strings in Javascript must have an encoding format associated. This way, when decoding `/ipns/BINARY_PEER_ID` to the topic string and sending the message to a go peer, it cannot parse the topic correctly.

Considering all this and that IPNS over Pubsub is still an experimental feature, I have been talking with @stebalien about changing the topic format to a standard encoding, such as `base64` for both implementations.

Stebalien:
So, some history: we *used* to use `/ipns/Qm...` but I changed it to `/ipns/BINARY_ID` to match what we do in the DHT. This allows us to treat pubsub as just another (kind of funky) record store (see https://github.com/libp2p/go-libp2p-pubsub-router). This, in turn, means we don't need to duplicate all the IPNS logic for both the DHT and Pubsub (or any other "record stores" we make in the future.

However, given that record-store keys are binary (arguably, they should also just be strings...), we could use `/rs/base32(key)` (or base64url). In this case, that would be `/record/$base64url(/ipns/BINARY_KEY)`.

However, I'd also like to consider just making pubsub topics arbitrary binary strings. We already do this for the DHT and forcing users to use UTF-8 for pubsub topics will likely cause issues like this in the future as well (some implementations will enforce it, others won't).

Stebalien:
I've hashed this out with @vyzo on IRC. He agrees we should probably be nice and use UTF8.

So, the only remaining question is: How? Personally, I'd vote for either `/record/base32(...)` or `/record/base64url(...)`. Thoughts?

vasco-santos:
Thanks for your analysis @Stebalien .

I would go with `/record/base64url(/ipns/BINARY_KEY)` since it will need fewer bytes and the topic will go with every single message. In addition, with `ipns.pubsub.subs` we can get the decoded topic.

Stebalien:
@vyzo suggested base32 to be consistent with everything else we do. However, I really don't mind either way. To get this ball rolling: https://github.com/libp2p/go-libp2p-pubsub-router/pull/17

magik6k:
I'd use base64 since this is internal, and we send these topics to all peers we are pubsubbing with, so b64 saves us a few bytes each time we connect to someone.

vasco-santos:
Interop between `go-ipfs` and `js-ipfs` is green locally using:

JS: [ipfs/js-ipfs#1559](https://github.com/ipfs/js-ipfs/pull/1559) with [ipfs/js-datastore-pubsub/pull/9](https://github.com/ipfs/js-datastore-pubsub/pull/9) for fixing the interop

Go: `go-ipfs#master` with [libp2p/go-libp2p-pubsub-router#17](https://github.com/libp2p/go-libp2p-pubsub-router/pull/17) (binary `go-ipfs` can be obtained with:  https://ipfs.io/ipfs/QmXQDCKdsEaeDyejWn2ziXfTSjBd5YEZ6zoHTdrpMJWyvf)

@Stebalien do you have any eta for the new `go-ipfs` release with this? 

@alanshaw maybe we can add this to [v0.34.0 RELEASE](https://github.com/ipfs/js-ipfs/issues/1721) 🚀

test/utils/wait-for.js
```diff
@@ -0,0 +1,24 @@
...
+  const self = setInterval(() => {
+    if (predicate()) {
+      clearInterval(self)
+      return callback()
+    }
+    if (Date.now() > ttl) {
+      clearInterval(self)
+      return callback(new Error('waitFor time expired'))
+    }
+  }, 500)
```
alanshaw: Personally I'd dial this down to 50ms or less!

test/ipns-pubsub.js
```diff
@@ -0,0 +1,172 @@
...
+
+      done()
+    })
+  })
+
+  // Get node ids
+  before(function (done) {
+    this.timeout(100 * 1000)
+
+    parallel([
```
alanshaw: Suggestion: `async/each`

test/ipns-pubsub.js
```diff
@@ -0,0 +1,172 @@
...
+      expect(err).to.not.exist()
+      expect(ids).to.exist()
+      expect(ids[0].id).to.exist()
+      expect(ids[1].id).to.exist()
+
+      nodeAId = ids[0]
+      nodeBId = ids[1]
+
+      nodes[0].api.swarm.connect(ids[1].addresses[0], () => {
+        setTimeout(done, 60000) // wait for republish as we can receive the republish message first
```
alanshaw: This is such a long wait, it would be great if this was somehow configurable...

@Stebalien is it possible in go-ipfs?

@vasco-santos perhaps add a `console.log` line here to inform the user this pause is happening...?

test/ipns-pubsub.js
```diff
@@ -0,0 +1,172 @@
...
+    ], (err, ids) => {
+      expect(err).to.not.exist()
+      expect(ids).to.exist()
+      expect(ids[0].id).to.exist()
+      expect(ids[1].id).to.exist()
+
+      nodeAId = ids[0]
+      nodeBId = ids[1]
+
+      nodes[0].api.swarm.connect(ids[1].addresses[0], () => {
```
alanshaw: Need to handle the error here.

test/ipns-pubsub.js
```diff
@@ -0,0 +1,172 @@
...
+const subscribeToReceiveByPubsub = (nodeA, nodeB, id, callback) => {
+  let subscribed = false
+  function checkMessage (msg) {
+    subscribed = true
+  }
+
+  const keys = ipns.getIdKeys(fromB58String(id))
+  const topic = `${namespace}${base64url.encode(keys.routingKey.toBuffer())}`
+
+  nodeB.api.name.resolve(id, () => {
```
alanshaw: This needs some explanation. What happens here? `nodeA` has not yet published anything for `self` - does it error or return `null`? We should handle the error at least.

test/ipns-pubsub.js
```diff
@@ -0,0 +1,172 @@
...
+    series([
+      (cb) => waitForPeerToSubscribe(nodeB.api, topic, cb),
+      (cb) => nodeB.api.pubsub.subscribe(topic, checkMessage, cb),
+      (cb) => nodeA.api.name.publish(ipfsRef, { resolve: false }, cb),
+      (cb) => nodeA.api.name.resolve(id, cb),
```
alanshaw: Why does `nodeA` need to `resolve` what it has just published?

test/ipns-pubsub.js
```diff
@@ -0,0 +1,172 @@
...
+      (cb) => waitFor(() => subscribed === true, (50 * 1000), cb),
+      (cb) => nodeB.api.name.resolve(id, cb)
+    ], (err, res) => {
+      expect(err).to.not.exist()
+      expect(res).to.exist()
+
+      expect(res[2].name).to.equal(id) // Published to Node A ID
+      expect(res[3]).to.equal(ipfsRef) // TODO: remove path once not using proc daemon
```
alanshaw: Has this `TODO` been done?

test/ipns-pubsub.js
vasco-santos: `go-ipfs` already accepts the `republish` configuration through the node config. For JS Land, it is one of the things that I want to integrate during this quarter.

Meanwhile, I can add a `console.log` here!

test/ipns-pubsub.js
vasco-santos: I think it is better to use `async/parallel` here, since using `async/each` here would require an extra array to keep the data with all the ids, which I feel unnecessary just for 2 elements.

test/ipns-pubsub.js
vasco-santos: Oh wait, `go-ipfs` allows the `RepublishPeriod` and `RecordLifetime`, but not the time for the first `InitialRebroadcastDelay`, afaik!

test/ipns-pubsub.js
vasco-santos: Does not need that resolve in fact! 

I was ensuring that both `resolve` have the same value, but I think it is better to remove it, in order to decrease the duration of the tests!

daviddias:
@vasco-santos mind rebasing master onto this PR and test it with js-ipfs master?

alanshaw:
@vasco-santos what's the status of this? Can we get this merged plz 🙏?

vasco-santos:
Subscribe in `js` and Publishing in `go` is currently failing. It seems that js node is not receiving the pubsub message once `go` publishes.

I just tested this using the latest `js-ipfs` and `go-ipfs` CLIs and everything worked fine, so I will debug this once I have some time

vasco-santos:
@alanshaw it is 💚 🎊

PR #53, opened 2019-02-12
Chore/move travis

hugomrdias:


daviddias:
Seems that there is a mismatch in deps/apis

image

But PR is good. Thank you for migrating this to Travis!! :D

PR #56, opened 2019-03-13
test: turn off discovery to make tests run faster

achingbrain:
Similar to ipfs/js-ipfs-http-client#942 - in my testing using the `test` profile for go and turning off mdns/webrtcstar decreases the time it takes to run the files tests (at least) from 15-45s to 7s


alanshaw:
Firefox failures seem unrelated, tests are passing in all other environments.

PR #59, opened 2019-03-29
test: improve exchange files

jacobheun:
This fixes an issue with relay configuration for the upcoming js-ipfs release.

Also: 
* timeouts were set way too high on the circuit tests. They shouldn't take that long, and its better they fail faster.
* windows cache isnt reliable on Travis right now, so I disabled that
* Exchange files timeouts were set at 15mins. Travis errors after 10mins of inactivity, so I dropped it down to 8mins.

### Exchange files
I refactored the exchange files tests to make them less prone to accidentally communicating with each other. While this is valuable scenario testing, it detracts from the goal of exchange files. I think there is value in having tests that exercise large data transfers with several nodes connected, but those should be separate tests. This will help prevent things like the dht from communicating with other nodes in the test suite without needing to disable them.

This does have a slight impact on startup times for the tests as the nodes are started and shutdown for each exchange files iteration, but overall it should help reduce time by avoiding unwanted network traffic.


alanshaw:
LGTM and tests passing locally

PR #60, opened 2019-04-04
feat: test against js-ipfs 0.35

alanshaw:


alanshaw:
@vasco-santos I'm getting this failure consistently, are you able to take a look?:

```console
  1) ipns locally using the same repo across implementations
       should publish an ipns record to a go daemon and resolve it using a js daemon through the reuse of the same repo:
     Uncaught AssertionError: expected [Error: record requested was not found for QmWyBY5suLNCtjXQA4p4cEtZMqctFWtfkYxWdwMweFC6Tr (/ipns/ �9B�`A\˹*�?�`g����pV}�U}g�) in the network] to not exist
      at resolverDaemon.api.name.resolve (test/ipns.js:68:28)
      at send (node_modules/ipfs-http-client/src/utils/send-request.js:215:16)
      at f (node_modules/once/once.js:25:25)
      at streamToJsonValue (node_modules/ipfs-http-client/src/utils/send-request.js:29:5)
      at streamToValue (node_modules/ipfs-http-client/src/utils/stream-to-json-value.js:30:5)
      at concat (node_modules/ipfs-http-client/src/utils/stream-to-value.js:12:22)
      at ConcatStream.<anonymous> (node_modules/ipfs-http-client/node_modules/concat-stream/index.js:32:43)
      at finishMaybe (node_modules/ipfs-http-client/node_modules/readable-stream/lib/_stream_writable.js:620:14)
      at afterWrite (node_modules/ipfs-http-client/node_modules/readable-stream/lib/_stream_writable.js:466:3)
      at processTicksAndRejections (internal/process/next_tick.js:76:17)
```

vasco-santos:
Something in `js-ipfs`  `0.34.4` -> `0.35.0-pre.0` lead to this failure. I will try to figure out the problem here.

vasco-santos:
@alanshaw figured out what the problem is.

There are two problems that make this test failing. First, the `ipfsd-ctl` is not dealing correctly with the `--offline` option. I will have a look on that. However, there is also an issue with `js-ipfs` that I fixed on [ipfs/js-ipfs#1988](https://github.com/ipfs/js-ipfs/pull/1988).

alanshaw:
@vasco-santos I've updated the IPFS dep - are we still waiting on a PR to `ipfsd-ctl`? I'd really like to get this merged today.

vasco-santos:
No, the problem was not with `ipfsd-ctl`. After debugging this:

> In the context of [ipfs/interop#60#issuecomment-482152895](https://github.com/ipfs/interop/pull/60#issuecomment-482152895), the offline flag should be provided through the daemon start, as this is is not a disposable daemon. Only non disposable daemons start can start automatically with the args provided in spawn.

So, the "problem" was that the args passed when spawning a daemon are not passed to `daemon.start`. Thinking about it, I think they should be and I will open an issue on the repo to discuss this with @hugomrdias . Anyway, we do not need to block this by that.

alanshaw:
Ok this is looking GTG! Firefox failure is a known issue.

PR #61, opened 2019-04-12
fix: ipns use offline on daemon start

vasco-santos:
In the context of [ipfs/interop#60#issuecomment-482152895](https://github.com/ipfs/interop/pull/60#issuecomment-482152895), the `offline` flag should be provided through the daemon `start`, as this is is not a disposable daemon. Only non disposable daemons start can start automatically with the `args` provided in `spawn`.


alanshaw:
This was merged with https://github.com/ipfs/interop/pull/60

PR #62, opened 2019-04-15
chore: disable firefox ci

vasco-santos:
We currently have some inconsistencies in the circuit relay tests between chrome and firefox tests. This way, we will disable the firefox job and track it in a new issue-

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,11 +32,5 @@ jobs:
         chrome: stable
       script: npx aegir test -t browser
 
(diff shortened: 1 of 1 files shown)


jacobheun:
This is fixed with the latest patch to js-libp2p-switch. I cleared the cache in Travis CI and reran the firefox build, it's all 💚's now. https://travis-ci.com/ipfs/interop/builds/108347555

Issue #16, opened 2018-03-07
circuit interop - `browser-browser-go` failing


dryajov:
When connection to a `go` node over a `browser` relay from another `browser` node, the relay fails. The two browser nodes are connected over `websocket-star`.

cc @vyzo

dryajov:
Here is a pcap capture from the establish relay step - [browser-browser-go.pcapng.zip](https://github.com/ipfs/interop/files/1791099/browser-browser-go.pcapng.zip)

diasdavid:
Was this fixed with https://github.com/libp2p/go-libp2p-circuit/pull/30 ?

dryajov:
Yep, it's taken care of there, but it's not going to be released until later, so perhaps it's worth keeping this around to track when that fix is finally released?

diasdavid:
@Stebalien refresh my memory, were the mplex + circuit relay fixes released with 0.4.15?

Stebalien:
No, they're in the latest master.

diasdavid:
One daaaaaaay 🎶

Stebalien:
@dryajov this appears to work. Has this been fixed?

Stebalien:
Nevermind...

PR #64, opened 2019-04-16
fix: enable browser-browser-go circuit test

jacobheun:
This reenables the `browser-browser-go` circuit test which is working.

fixes https://github.com/ipfs/interop/issues/16


jacobheun:
The circuit tests are passing, but ipns still appears to be flakey. @vasco-santos any idea why?

vasco-santos:
wow, that is a new failure. Let me have a look

vasco-santos:
So `go-ipfs` now have a new property for publishing names offline: https://github.com/ipfs/go-ipfs/blob/master/core/commands/name/publish.go#L24

I just added a commit passing that and I will open an issue in `js-ipfs` for adding that property.

vasco-santos:
We got a cid test failure on windows. I restarted the job

PR #66, opened 2019-04-29
fix: pubsub tests

jacobheun:
Looking into `pull-mplex` updates in js-ipfs I noticed the pubsub tests here needed improvement. For example, JS -> JS and go -> go just use the same node for publishing and subscribing. I refactored the tests to spin up unique nodes for each suite.

This exposed a current issue with go -> js pubsub failing.  I am looking into this.

- [x] depends on https://github.com/dignifiedquire/pull-length-prefixed/pull/23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ os:
   - linux
   - osx
 
-script: npx nyc -s npm run test:node -- --bail
+script: npm i pull-length-prefixed@dignifiedquire/pull-length-prefixed#fix/empty-streams  && npx nyc -s npm run test:node -- --bail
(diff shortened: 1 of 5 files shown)


alanshaw:
LGTM, windows failure is persistent though. Looks vaguely unrelated but `spawnInitAndStartJsDaemon` _is_ used in the CID version agnostic tests (and has been changed here).

jacobheun:
@alanshaw it looks like this is flakey timeout issue. I reran windows and it passed this time. I think I've seen the `before` fail previously. I bumped the timeout to help avoid the flakey failure in the future.

PR #70, opened 2019-05-30
test: disable mdns in exchange files

jacobheun:
The tests are failing against the latest go-ipfs rc candidate (0.4.21-rc.3), https://github.com/ipfs/go-ipfs/issues/6389. This is due to an issue with go announcing its address over MDNS with port 4001 even though the port is actually different. Since the MDNS discovery occurs before dial in the test and fails, the connect attempt will subsequently fail due to JS doing a blacklisting backoff on failed connection attempts. 

While this doesn't solve the underlying issue, we shouldn't have MDNS turned on for these tests. My recommendation is:
> I'd recommend we disable MDNS in the tests for now and file an issue for fixing MDNS so this isn't blocking the release as it's not really a critical issue. In reality we should just move to the new MDNS spec for both implementations and not waste time on a fix for this. The default ipfs daemon runs on port 4001 so this shouldn't be an issue in most cases, and the interop fix was a recent update for js.

### DIFF
--- a/test/exchange-files.js
+++ b/test/exchange-files.js
@@ -100,11 +100,38 @@ const min = 60 * 1000
 const timeout = isCi ? 8 * min : 5 * min
 
 function createJs (cb) {
-  jsDf.spawn({ type: 'js', initOptions: { bits: 512 }, config: { Bootstrap: [] } }, cb)
+  jsDf.spawn({
+    type: 'js',
+    initOptions: { bits: 512 },
+    config: {
+      Bootstrap: [],
+      Discovery: {
+        MDNS: {
+          Enabled: false
+        },
+        webRTCStar: {
+          Enabled: false
+        }
+      }
+    }
+  }, cb)
 }
 
 function createGo (cb) {
-  goDf.spawn({ initOptions: { bits: 1024 }, config: { Bootstrap: [] } }, cb)
+  goDf.spawn({
+    initOptions: { bits: 1024 },
+    config: {
+      Bootstrap: [],
+      Discovery: {
+        MDNS: {
+          Enabled: false
+        },
+        webRTCStar: {
+          Enabled: false
+        }
+      }
+    }
+  }, cb)
 }
 
 describe('exchange files', () => {
--- a/test/exchange-files.js
+++ b/test/exchange-files.js
@@ -108,9 +108,6 @@ function createJs (cb) {
       Discovery: {
         MDNS: {
           Enabled: false
-        },
-        webRTCStar: {
-          Enabled: false
         }
       }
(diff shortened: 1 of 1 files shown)


Stebalien:
> While this doesn't solve the underlying issue, we shouldn't have MDNS turned on for these tests.

We should also look into fixing our backoff logic. Ideally:

1. We'd backoff specific addresses in the swarm.
2. We'd backoff peer routing in the DHT.

That way, if we learn _new_ addresses, we'll dial those.

test/exchange-files.js
```diff
@@ -100,11 +100,38 @@ const min = 60 * 1000
...
-  goDf.spawn({ initOptions: { bits: 1024 }, config: { Bootstrap: [] } }, cb)
+  goDf.spawn({
+    initOptions: { bits: 1024 },
+    config: {
+      Bootstrap: [],
+      Discovery: {
+        MDNS: {
+          Enabled: false
+        },
+        webRTCStar: {
```
Stebalien: go-ipfs will just ignore this, IIRC.

jacobheun: correct, and we don't really need it for js for this, lazy copy paste on my part, i've removed it from both.

jacobheun:
> We'd backoff specific addresses in the swarm.

That's a good point, we're doing it at the peer level now and not doing any reseting when addresses are updated. This shouldn't take much to transition to for js.

> We'd backoff peer routing in the DHT.

What were your thoughts for this, not sure I follow?

Stebalien:
My thinking is that we should completely drop the per-peer logic. The point of the backoff is to avoid repeatedly doing something that fails over and over.

So, we can:

1. Rate limit finding peers in the DHT. If we've searched for a peer within the last N minutes (and we didn't _cancel_ the search half-way through), return the cached results.
2. Add each individual multiaddr to a blacklist (per peer) instead of blocking the entire peer. That way, if we get new addresses, we can skip the ones that don't work.

---

But you're right, the simple solution is to just remove the backoff when we learn a new address. In go-ipfs, we reset the backoff when the user calls the `ipfs connect` command manually.

jacobheun:
Ah, yeah caching the DHT query would be helpful. I created an issue to transition to a multiaddr based backoff for js, https://github.com/libp2p/js-libp2p-switch/issues/339. If it becomes a more prevalent issue before I can get it fixed we could replicate the same backoff reset behavior of go-ipfs pretty easily for js-ipfs.

jacobheun:
The test failures are unrelated to this and exist in master. I created an issue to track that failure, https://github.com/ipfs/interop/issues/71.

Issue #71, opened 2019-05-31
IPNS Pubsub tests failing on master for linux


jacobheun:
The test is only failing on linux. @vasco-santos any thoughts on what's going on here?

```log
ipns-pubsub
       should publish the received record to a go node and a js subscriber should receive it:
     Uncaught AssertionError: expected [Error: waitFor time expired] to not exist
      at series (test/ipns-pubsub.js:145:26)
      at /home/travis/build/ipfs/interop/node_modules/async/internal/parallel.js:39:9
      at /home/travis/build/ipfs/interop/node_modules/async/internal/once.js:12:16
      at iterateeCallback (node_modules/async/internal/eachOfLimit.js:45:17)
      at /home/travis/build/ipfs/interop/node_modules/async/internal/onlyOnce.js:12:16
      at /home/travis/build/ipfs/interop/node_modules/async/internal/parallel.js:36:13
      at Timeout.setInterval (test/utils/wait-for.js:21:14)
```

https://travis-ci.com/ipfs/interop/jobs/202899201

vasco-santos:
I will have a look

vasco-santos:
I have not been able to reproduce with a linux machine. I will look at installing the linux os running in CI

vasco-santos:
So, I was finally able to debug the problem until finding out what hit these tests.

The PR with tests for adding and modifying mfs hamt shards [ipfs/interop#54](https://github.com/ipfs/interop/pull/54) bumped the `cid` module from version `~0.5.7` to `~0.7.1`, which was the dependency that broke this test.

I looked into the progress of the `cid` module from `~0.5.7` version and the commit which broke this test was [multiformats/js-cid/commit/0e94b5531c98b4bde87eef9ee7b76e06f8a4f6a7](https://github.com/multiformats/js-cid/commit/0e94b5531c98b4bde87eef9ee7b76e06f8a4f6a7) in [multiformats/js-cid#77](https://github.com/multiformats/js-cid/pull/77).

I tested removing those validations with [multiformats/js-cid/commit/32eaf3fea10bb13eec7c92e64eb90a02342dd73c](https://github.com/multiformats/js-cid/commit/32eaf3fea10bb13eec7c92e64eb90a02342dd73c) and installed this commit with interop   [ipfs/interop#72](https://github.com/ipfs/interop/pull/72) and the tests get green.

[...]

PR #73, opened 2019-06-05
chore: update dependencies

vasco-santos:
This PR addresses [ipfs/interop#71](https://github.com/ipfs/interop/issues/71). IPNS Pubsub tests were failing on master for linux, as a result of `cids` being outdated on `ipfsd-ctl`

### DIFF
--- a/test/files.js
+++ b/test/files.js
@@ -23,14 +23,14 @@ class ExpectedError extends Error {
 function checkNodeTypes (daemon, file) {
   return daemon.api.object.get(file.hash)
     .then(node => {
-      const meta = UnixFs.unmarshal(node.data)
+      const meta = UnixFs.unmarshal(node.Data)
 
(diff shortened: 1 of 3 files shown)


test/files.js
```diff
@@ -23,14 +23,14 @@ class ExpectedError extends Error {
 function checkNodeTypes (daemon, file) {
   return daemon.api.object.get(file.hash)
     .then(node => {
-      const meta = UnixFs.unmarshal(node.data)
+      const meta = UnixFs.unmarshal(node.Data)
```
daviddias: @vmx this looks like an API breaking change. Was it treated like so?

vmx: Yes.

vmx: https://github.com/ipld/js-ipld-dag-pb/releases/tag/v0.16.0

PR #81, opened 2019-07-19
fix: ipns pubsub race condition

### LINKED ISSUE #80: potential race condition on ipns-pubsub tests
https://github.com/ipfs/interop/blob/master/test/ipns-pubsub.js#L151-L153

From IRC

```
daviddias
https://github.com/ipfs/interop/blob/master/test/ipns-pubsub.js#L151-L153 could be a race condition
The subscribe does not necessarily wait to confirm that nodeA received the message
the test should be updated to have another waitFor until nodeA shows that it knows of nodeB subscription
```

This might cause one of the errors @Stebalien is seeing at https://circleci.com/gh/ipfs/go-ipfs/18357 //cc @vasco-santos

vasco-santos:
Added verification for guaranteeing that the peer publishing a record knows about the subscription of the peer resolving the record.

Closes #80

### DIFF
--- a/test/ipns-pubsub.js
+++ b/test/ipns-pubsub.js
@@ -123,45 +123,74 @@ describe('ipns-pubsub', () => {
   it('should publish the received record to a go node and a js subscriber should receive it', function (done) {
     this.timeout(300 * 1000)
 
-    subscribeToReceiveByPubsub(nodes[0], nodes[1], nodeAId.id, done)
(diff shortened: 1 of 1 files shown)


hugomrdias:
@Stebalien this makes circle ci pass ?

Stebalien:
I have no idea. However, it looks like it solves a real issue. My plan is to merge and then re-run CI.

hugomrdias:
lets do it then

PR #85, opened 2019-09-20
chore: remove unused deps

hugomrdias:


achingbrain:
Looks like CI is picking up linting errors from the stealth standard upgrade in Aegir.

Issue #84, opened 2019-09-11
Convert tests to use `async/await`


achingbrain:
As part of the [great migration to async iterators](https://github.com/ipfs/js-ipfs/issues/1670) we need to convert our tests to use async/await.  This is often best done separately from converting the tested code from callbacks to promises.

This repo's tests need converting.

PedroMiguelSS:
This issue will be tackled following these steps:
- [ ] Update `ipfsd-ctl` version to support promises and make the necessary changes to make sure the tests can run successfully
- [ ] Change all the tests to use the same utils file (`daemon.js`) to init and spawn the daemons
- [ ] Update the util file to be consistent with what's being done on `js-ipfs-http-client` tests
- [x] Change the tests to use async/await

achingbrain:
Fixed by #87

PR #86, opened 2019-09-23
chore: update ipfsd-ctl version and use daemon.js util file

PedroMiguelSS:
Completes phase **1** and **2** to tackle #84 as referenced [here](https://github.com/ipfs/interop/issues/84#issuecomment-536503271).

Some tests were skipped due to an inconsistency: the same content is getting different hashes when added to a `go daemon` and a `js daemon`. The skipped tests are:

### `files.js` file

- **trickle DAGs**
trickle

- **rabin chunker**
rabin

- **rabin chunker small chunks**
rabin small

⚠️These tests must be unskipped as soon as the problem gets solved.


test/pin.js
```diff
@@ -22,140 +22,139 @@ describe('pin', function () {
...
+  async function spawnAndStart (type, repoPath = utils.tmpPath()) {
+    const daemonOptions = {
+      repoPath,
+      disposable: false
+    }
+
+    const daemon = await (type === 'go' ? spawnInitAndStartGoDaemon(daemonOptions) : spawnInitAndStartJsDaemon(daemonOptions))
+    daemons.push(daemon)
+
+    if (daemon.initialized) {
```
vasco-santos: If we do `spawnInitAndStartGoDaemon` / `spawnInitAndStartJsDaemon`, this logic for init/start should just go away?

test/repo.js
```diff
@@ -6,118 +6,90 @@ const dirtyChai = require('dirty-chai')
...
 
-const DaemonFactory = require('ipfsd-ctl')
-const goDf = DaemonFactory.create()
-const jsDf = DaemonFactory.create({ type: 'js' })
+const { spawnInitAndStartGoDaemon, spawnInitAndStartJsDaemon } = require('./utils/daemon')
+const timeout = require('./utils/timeout')
 
-function catAndCheck (api, hash, data, callback) {
+async function catAndCheck (api, hash, data) {
   api.cat(hash, (err, fileData) => {
```
vasco-santos: this should be removed, right?

test/repo.js
```diff
@@ -6,118 +6,90 @@ const dirtyChai = require('dirty-chai')
...
+    const res = await goDaemon.api.add(data)
+    const hash = res[0].hash
+
+    await catAndCheck(goDaemon.api, hash, data)
+    await goDaemon.stop()
+
+    const jsDaemon = await spawnInitAndStartJsDaemon({
+      repoPath: dir,
+      disposable: false,
+      initOptions: { bits: 512 }
```
vasco-santos: remove?

achingbrain:
I'm a little concerned by the skipped tests around different hashes.

These tests are run on the current release & master of js-IPFS and all pass - why has converting them to async/await made them fail?  Or is it something to do with a new dep?

PedroMiguelSS:
I've debugged a little bit more and I found out that the problem on both `remove a child shared by multiple pins` and `print same pins` tests (`pin.js` file) was related to the options that were being passed to the `spawn()` method. The default config had `sharding` option enabled and this was causing the problem on the latest versions of `ipfsd-ctl` (`v0.47.1`). However, it's fixed and these 2 tests are passing now 🎉

However, I still have 3 skipped tests on `files.js` file as you can see in the description of this PR. I found out that this problem only occurs exactly on `v0.46.1` of `ipfsd-ctl`. Why? There is one [commit](https://github.com/ipfs/js-ipfsd-ctl/commit/2f61eef416bbed360e724f562da7bd119d9259d7) that updates several dependencies. One of the updated dependencies is the `ipfs-http-client` and it was updated from `v34.0.0` to `v35.1.0`. Thus, there was a breaking change and the `add` method was changed as we can see [here](https://github.com/ipfs/js-ipfs-http-client/commit/9616aed5ece058a3c98d12aa2dbe2999d08acb1f).

Can I get some help here?
@achingbrain @alanshaw @hugomrdias

test/cid-version-agnostic.js
```diff
@@ -20,8 +22,8 @@ describe('CID version agnostic', () => {
     this.timeout(50 * 1000)
 
     const [js0, js1, go0, go1] = await Promise.all([
-      spawnInitAndStartJsDaemon(),
-      spawnInitAndStartJsDaemon(),
+      spawnInitAndStartJsDaemon(jsDaemonOptions),
+      spawnInitAndStartJsDaemon(jsDaemonOptions),
```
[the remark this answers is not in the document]
PedroMiguelSS: Setting the `Bootstrap` list to empty is speeding up the test execution. I can remove this option but I'll need to add a custom `timeout` to the test. Which approach do you prefer?

test/files.js
```diff
@@ -299,7 +292,10 @@ describe('files', function () {
       )
     })
 
-    it('trickle DAGs', () => {
+    // Somehow, the hashes for the same content are not equal
+    // This problem was also detected on
+    // both 'remove a child shared by multiple pins' test and 'print same pins'
+    it.skip('trickle DAGs', () => {
```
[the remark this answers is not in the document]
PedroMiguelSS: This is currently fixed 👍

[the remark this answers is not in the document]
hugomrdias: if its fixed we should remove the skip

test/utils/timeout.js
```diff
@@ -0,0 +1,3 @@
+'use strict'
+
+module.exports = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
```
hugomrdias: use `delay` package

test/exchange-files.js
```diff
@@ -99,41 +97,17 @@ if (isCi) {
...
-    config: {
-      Bootstrap: [],
-      Discovery: {
-        MDNS: {
-          Enabled: false
-        }
-      }
-    }
-  }, cb)
+const jsDaemonOptions = {
```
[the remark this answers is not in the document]
PedroMiguelSS: I'll remove `initOptions` and keep the `Bootstrap` list

test/repo.js
PedroMiguelSS: Right!

test/files.js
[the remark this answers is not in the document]
PedroMiguelSS: The fixed ones are not these 3 tests. The fixed are: [`remove a child shared by multiple pins`](https://github.com/ipfs/interop/blob/chore/update-ipfsd-ctl/test/pin.js#L115) and [`print same pins`](https://github.com/ipfs/interop/blob/chore/update-ipfsd-ctl/test/pin.js#L141) and they're already unskipped.

test/utils/daemon.js
```diff
@@ -11,49 +11,22 @@ const spawnInitAndStartDaemon = (factory, options) => {
...
-      }
-
-      resolve()
-    })
-  })
-}
+const stopDaemon = (daemon) => daemon.stop()
 
 module.exports = {
   spawnInitAndStartDaemon,
```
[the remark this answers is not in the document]
PedroMiguelSS: I agree!

test/pin.js
PedroMiguelSS: I'll rename these functions' names to `spawnGoDaemon` and `spawnJsDaemon` and this way everything is more intuitive.

hugomrdias:
@PedroMiguelSS check whats happening with the CI

PedroMiguelSS:
@hugomrdias, issue: #88

hugomrdias:
@achingbrain we still have a couple of timeout error coming from the go daemon on windows by this should be good to go, can you review it again pls

test/cid-version-agnostic.js
```diff
@@ -43,99 +44,98 @@ describe('CID version agnostic', () => {
   })
 
   after(function () {
-    this.timeout(30 * 1000)
-    return Promise.all(Object.values(daemons).map(stopDaemon))
+    return Promise.all(Object.values(daemons).map((daemon) => daemon.stop()))
   })
 
-  it('should add v0 and cat v1 (go0 -> go0)', async () => {
+  it('should add v0 and cat v1 (go0 -> go0)', async function () {
```
achingbrain: Why did this change from an arrow function?

test/circuit/browser.js
```diff
@@ -19,133 +19,124 @@ const base = '/ip4/127.0.0.1/tcp/0'
...
   'browser-js-go': {
-    create: (callback) => series([
-      (cb) => createProc([], cb),
-      (cb) => createJs([`${base}/ws`], cb),
-      (cb) => createJs([`${base}/ws`], cb)
-    ], callback),
+    create: () => Promise.all([
+      createProc([]),
+      createJs([`${base}/ws`]),
+      createGo([`${base}/ws`])
```
achingbrain: Nice catch..

test/files.js
```diff
@@ -299,7 +306,10 @@ describe('files', function () {
...
```
achingbrain: Does this still need to be skipped?

test/ipns-pubsub.js
```diff
@@ -10,105 +10,54 @@ const { fromB58String } = require('multihashes')
...
-      expect(ids[1].id).to.exist()
+  before(async function () {
+    nodesIds = await Promise.all([
+      nodes[0].api.id(),
+      nodes[1].api.id()
+    ])
 
-      nodeAId = ids[0]
-      nodeBId = ids[1]
+    expect(nodesIds[0]).to.exist()
```
achingbrain: You can collapse the assertions as:

```javascript
expect(nodesIds).to.have.nested.property('[0].id').that.is.ok()
expect(nodesIds).to.have.nested.property('[1].id').that.is.ok()
```

or

```javascript
expect(nodesIds).to.all.have.property('id').that.is.ok()
```

test/ipns-pubsub.js
```diff
@@ -10,105 +10,54 @@ const { fromB58String } = require('multihashes')
...
-        expect(err).to.not.exist()
+    await nodes[0].api.swarm.connect(nodesIds[1].addresses[0])
 
-        // eslint-disable-next-line no-console
-        console.log('wait for republish as we can receive the republish message first')
-        setTimeout(done, 60000) // wait for republish as we can receive the republish message first
-      })
-    })
+    console.log('wait for republish as we can receive the republish message first') // eslint-disable-line
+    return new Promise((resolve) => setTimeout(resolve, 60000))
```
achingbrain: Use the `delay` module here?

test/pin.js
```diff
@@ -22,140 +22,135 @@ describe('pin', function () {
...
-        .then(([goPins, jsPins]) => {
-          expect(goPins.length).to.be.gt(0)
-          expect(goPins).to.deep.include.members(jsPins)
-          expect(jsPins).to.deep.include.members(goPins)
+      const [goPins, jsPins] = await withDaemons(pipeline)
 
-          const dirPin = goPins.find(pin => pin.hash === jupiterDir.hash)
-          expect(dirPin.type).to.eql('indirect')
-        })
+      expect(goPins.length).to.be.gt(0)
```
achingbrain: Avoids the possibility of an unhelpful error message if `goPins` is undefined:

```javascript
expect(goPins).to.have.property('length').that.is.gt(0)
```

achingbrain:
Tiny nits, just needs the added `.skip`s removing really.

test/cid-version-agnostic.js
PedroMiguelSS: I added a custom `this.timeout()` to these tests. Then I removed it and I forgot to change the function back to an arrow function.

test/files.js
PedroMiguelSS: These 3 tests cannot be unskipped as they are still failing. Issue number #88 has further details about this problem.
Should we block this PR until we get this solved?

achingbrain: I'm taking a look at those failing tests

achingbrain: Though having said that I do seem to be getting different CIDs from js-IPFS master on the cli if the daemon is on or off. 😬

achingbrain: `ipfs-http-client` wasn't inspecting the right field for the `--trickle` arg - the fix is in https://github.com/ipfs/js-ipfs-http-client/pull/1129 and I'll backport it to `ipfs-http-client@38.0.x` when it's merged.

achingbrain: Urgh, no, it's more complicated that that, sorry - tired.  `ipfs-http-client` is doing the right thing - it takes a `trickle` key as part of it's options and sends that over http.  The cli in `ipfs` takes a `trickle` argument but turns it into `strategy: 'trickle'` (as that's what's required by the `unixfs-importer`) and passes that to `ipfs.add`* and the http api endpoint does the same thing.

The right thing to do is for `ipfs-http-client`,  the cli and the http api endpoint to all pass `trickle: true` to `ipfs.add` (as per the interface spec) and have *that* do the conversion to `strategy: 'trickle'`.

* = actually `ipfs._addAsyncIterator` but `ipfs.add` just passes it's options straight through so same same.

achingbrain: https://github.com/ipfs/js-ipfs/pull/2541

achingbrain:
I think this is good to go, will merge when CI passes.  Looks like some flaky tests in there.

achingbrain:
Famous last words, perhaps.

achingbrain:
I've replicated the failing tests locally, seems this is the fix: https://github.com/ipfs/js-ipfs-http-client/pull/1130

achingbrain:
Tests passing, hooray!

momack2:
CONGRATULATIONS ON THE AWESOME MERGE, @PedroMiguelSS and @achingbrain!!

PR #87, opened 2019-09-30
chore: update tests to async await

PedroMiguelSS:
This PR changes all the tests to use async/await. 
It completes phase number **4** to tackle #87 as referenced [here](https://github.com/ipfs/interop/issues/84#issuecomment-536503271).


hugomrdias:
CI still red and conflicts can you fix this so we can review ?

test/utils/wait-for.js
```diff
@@ -3,22 +3,19 @@
...
  * Wait for a condition to become true.  When its true, callback is called.
  */
-module.exports = (predicate, ttl, callback) => {
-  if (typeof ttl === 'function') {
-    callback = ttl
-    ttl = Date.now() + (10 * 1000)
-  } else {
-    ttl = Date.now() + ttl
-  }
+module.exports = (predicate, ttl) => {
```
vasco-santos:
```suggestion
module.exports = (predicate, ttl = 10e3) => {
```

test/ipns-pubsub.js
```diff
@@ -66,26 +70,23 @@ describe('ipns-pubsub', () => {
...
+    expect(state).to.exist()
+    expect(state.enabled).to.equal(true)
   })
 
-  it('should publish the received record to a go node and a js subscriber should receive it', function (done) {
+  it('should publish the received record to a go node and a js subscriber should receive it', async function () {
     this.timeout(300 * 1000)
 
-    subscribeToReceiveByPubsub(nodes[0], nodes[1], nodesIds[0].id, nodesIds[1].id, done)
+    await subscribeToReceiveByPubsub(nodes[0], nodes[1], nodesIds[0].id, nodesIds[1].id)
```
vasco-santos: return instead?

test/ipns-pubsub.js
```diff
@@ -107,64 +108,43 @@ const subscribeToReceiveByPubsub = (nodeA, nodeB, idA, idB, callback) => {
...
+    await waitForPeerToSubscribe(nodeB.api, topic)
+    await nodeB.api.pubsub.subscribe(topic, checkMessage)
+    await waitForNotificationOfSubscription(nodeA.api, topic, idB)
+    const res1 = await nodeA.api.name.publish(ipfsRef, { resolve: false })
+    await waitFor(() => subscribed === true, (50 * 1000))
+    const res2 = await nodeB.api.name.resolve(idA)
+
+    expect(res1.name).to.equal(idA) // Published to Node A ID
+    expect(res2).to.equal(ipfsRef)
+  }
```
vasco-santos: So, if `await nodeB.api.name.resolve(idA)` happens to run correctly, instead of failing, this test will pass. Please make a return by the end of catch + `thrown Error` after the catch

test/ipns-pubsub.js
```diff
@@ -107,64 +108,43 @@ const subscribeToReceiveByPubsub = (nodeA, nodeB, idA, idB, callback) => {
...
-
-      if (!res || !res.length || !res.includes(peerId)) {
-        return next(new Error('Could not find peer subscribing'))
-      }
-
-      return next(null)
-    })
-  }, callback)
+const waitForNotificationOfSubscription = async (daemon, topic, peerId) => {
+  await pRetry(async () => {
```
vasco-santos: return?

test/ipns-pubsub.js
```diff
@@ -107,64 +108,43 @@ const subscribeToReceiveByPubsub = (nodeA, nodeB, idA, idB, callback) => {
...
-
-      if (!res || !res.length || !res.includes(topic)) {
-        return next(new Error('Could not find subscription'))
-      }
-
-      return next(null, res[0])
-    })
-  }, callback)
+const waitForPeerToSubscribe = async (daemon, topic) => {
+  await pRetry(async () => {
```
vasco-santos: return?

test/utils/circuit.js
```diff
@@ -137,19 +131,12 @@ const getCircuitAddr = (addrs) => addrs
...
-    (cb) => nodeB.ipfsd.api.swarm.connect(getWsAddr(relay.addrs), cb),
-    // TODO: needed until https://github.com/ipfs/interop/issues/17 is resolved
-    (cb) => setTimeout(cb, timeout),
-    (cb) => nodeA.ipfsd.api.swarm.connect(getCircuitAddr(nodeB.addrs), cb)
-  ], callback)
+const connect = async (nodeA, nodeB, relay, timeoutMs = 1000) => {
+  await nodeA.ipfsd.api.swarm.connect(getWsAddr(relay.addrs))
+  await nodeB.ipfsd.api.swarm.connect(getWsAddr(relay.addrs))
+  // TODO: needed until https://github.com/ipfs/interop/issues/17 is resolved
+  await timeout(timeoutMs)
```
vasco-santos: should we use the delay package instead?

test/pubsub.js
```diff
@@ -53,40 +47,35 @@ describe('pubsub', function () {
...
 
-      before('spawn nodes', function () {
+      before('spawn nodes', async function () {
         this.timeout(timeout)
 
-        return Promise.all(tests[name].map(fn => fn()))
-          .then(nodes => {
-            [daemon1, daemon2] = nodes
-          })
+        const nodes = await Promise.all(tests[name].map(fn => fn()))
```
vasco-santos: destruct?

test/exchange-files.js
```diff
@@ -125,31 +122,27 @@ describe('exchange files', () => {
...
+          daemon2.api.id()
+        ])
+
+        id1 = ids[0]
+        id2 = ids[1]
+
+        await daemon1.api.swarm.connect(id2.addresses[0])
+        await daemon2.api.swarm.connect(id1.addresses[0])
+
+        const peers = await Promise.all([
```
achingbrain: This can also be destructured:

```javascript
const [ peers1, peers2 ] = await Promise.all([
  daemon1.api.swarm.peers(),
  daemon2.api.swarm.peers()
])

expect(peers1.map((p) => p.peer.toB58String())).to.include(id2.id)
expect(peers2.map((p) => p.peer.toB58String())).to.include(id1.id)
```

test/files.js
```diff
@@ -30,89 +30,82 @@ const jsOptions = {
...
 
-  return Promise.all(
-    ops
-  )
-    .then(results => {
-      expect(results.length).to.equal(ops.length)
+  const results = await Promise.all(ops)
 
-      const result = results.pop()
+  expect(results.length).to.equal(ops.length)
```
achingbrain:
```suggestion
  expect(results).to.have.lengthOf(ops.length)
```

test/files.js
```diff
@@ -30,89 +30,82 @@ const jsOptions = {
...
-      results.forEach(res => expect(res).to.deep.equal(result))
-    })
+  const result = results.pop()
+
+  results.forEach(res => expect(res).to.deep.equal(result))
 }
 
-const compareErrors = (expectedMessage, ...ops) => {
+const compareErrors = async (expectedMessage, ...ops) => {
   expect(ops.length).to.be.above(1)
```
achingbrain:
```suggestion
  expect(ops).to.have.property('length').that.is.above(1)
```

test/files.js
```diff
@@ -30,89 +30,82 @@ const jsOptions = {
...
     }))
-  )
-    .then(results => {
-      expect(results.length).to.equal(ops.length)
 
-      // all implementations should have similar error messages
-      results.forEach(res => {
-        expect(res.message.toLowerCase()).to.contain(expectedMessage.toLowerCase())
-      })
+  expect(results.length).to.equal(ops.length)
```
achingbrain:
```suggestion
  expect(results).to.have.lengthOf(ops.length)
```

test/files.js
```diff
@@ -30,89 +30,82 @@ const jsOptions = {
...
-      // all implementations should have the same error code
-      results.forEach(res => {
-        expect(res.code).to.equal(result.code)
-      })
-    })
+  const result = results.pop()
+
+  // all implementations should have the same error code
+  results.forEach(res => {
+    expect(res.code).to.equal(result.code)
```
achingbrain:
```suggestion
    expect(res).to.have.property('code', result.code)
```

test/ipns-pubsub.js
achingbrain: I think `await` here is fine since there's no need to return a value from this function.

test/ipns-pubsub.js
```diff
@@ -107,64 +108,43 @@ const subscribeToReceiveByPubsub = (nodeA, nodeB, idA, idB, callback) => {
   const topic = `${namespace}${base64url.encode(keys.routingKey.toBuffer())}`
 
   // try to resolve a unpublished record (will subscribe it)
-  nodeB.api.name.resolve(idA, (err) => {
+  try {
+    await nodeB.api.name.resolve(idA)
```
achingbrain: Instead of try/catch:

```suggestion
    await expect(nodeB.api.name.resolve(idA)).to.be.rejected()
```

The other assertions can then continue as normal without being in a `catch` block that may not get executed as @vasco-santos points out.

test/ipns-pubsub.js
achingbrain: See my comment about `expect(..).to.be.rejected()` above for how to not need to throw anything.

test/ipns-pubsub.js
achingbrain: See earlier comment, we don't need to return anything from this function so let's not.

test/utils/wait-for.js
```diff
@@ -3,22 +3,19 @@
...
-    if (predicate()) {
-      clearInterval(self)
-      return callback()
-    }
-    if (Date.now() > ttl) {
-      clearInterval(self)
-      return callback(new Error('waitFor time expired'))
-    }
-  }, 50)
+  return new Promise((resolve, reject) => {
```
achingbrain:
```javascript
'use strict'

const delay = require('delay')

module.exports = (predicate, ttl = 10e3, checkInterval = 50) => {
  const timeout = Date.now() + ttl

  while (true) {
    if (predicate()) {
      return
    }

    await delay(checkInterval)

    if (Date.now() > timeout) {
      throw new Error('waitFor time expired')
    }
  }
}
```

test/utils/p-retry.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
+
+const pRetry = require('p-retry')
+const timeout = require('./timeout')
+
+/*
+ * This is a wrapper for pRetry enabling an interval value
+ * i.e., the time to wait between retries, in milliseconds
+ */
```
achingbrain: Maybe add a note that we can remove this if https://github.com/sindresorhus/p-retry/issues/27 gets resolved?

test/ipns-pubsub.js
PedroMiguelSS: What about `expect.fail(...)`?

test/pubsub.js
PedroMiguelSS: Somehow I'm getting a problem while destructuring in some cases. This is a good example:

It works ✅
```js
const nodes = await Promise.all(tests[name].map(fn => fn()))

console.log(nodes)
// Logs:
// [(...)]
```

It doesn't work ❌
```js
[daemon1, daemon2] = await Promise.all(tests[name].map(fn => fn()))

console.log(daemon1)
// Logs:
// undefined
console.log(daemon2)
// Logs:
// undefined
```

Am I doing something wrong?

achingbrain: You need to add a semicolon between `this.timeout(timeout)` and the destructuring assignment.

E.g. this:

```javascript
this.timeout(timeout)

[daemon1, daemon2] = await Promise.all(tests[name].map(fn => fn()))
```

needs to become this:

```javascript
this.timeout(timeout);

[daemon1, daemon2] = await Promise.all(tests[name].map(fn => fn()))
```

otherwise the runtime thinks you're trying to do:

```javascript
this.timeout(timeout)[daemon1, daemon2] = await Promise.all(tests[name].map(fn => fn()))
```

and throws an error because `this.timeout()` doesn't return anything.

If it did it would try to assign an array with two members to the key `undefined` of the return value which is almost certainly not what you want:

```javascript

let val1, val2

async function main () {
  const val = {}

  function fn () {
    return val
  };

  fn()

  [val1, val2] = await Promise.all([1, 2])

  console.info(val)
  // prints "{ undefined: [ 1, 2 ] }"
}

main()
```

Anyway there's more info here: http://inimino.org/~inimino/blog/javascript_semicolons

package.json
```diff
@@ -69,6 +69,7 @@
     "mocha": "^5.2.0",
```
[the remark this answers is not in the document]
PedroMiguelSS: Done! 👍

test/ipns-pubsub.js
PedroMiguelSS: Cool stuff. I didn't know about that!

test/ipns-pubsub.js
PedroMiguelSS: I did change it to `expect(..).to.be.rejected()` as @achingbrain suggested.

PedroMiguelSS:
I've committed the requested changes. Can you review it again, please? @achingbrain @vasco-santos

daviddias:
@PedroMiguelSS see error's on CI: https://travis-ci.com/ipfs/interop/jobs/250668939#L329-L372 can you figure out why it is failing?

alanshaw:
🎉👏❤️

PR #90, opened 2019-10-22
refactor: make ipfs and http client peer dependencies

achingbrain:
By depending on a certain version of ipfs and the ipfs http client it means we can't run all the tests on other versions of ipfs/the http client as the `require`s resolve to the wrong version.

This PR makes them peer dependencies to allow the tester/testing environment to specify which versions of modules we want to test.


alanshaw:
cc @Stebalien this will effect go-ipfs CI

Stebalien:
So we'll have to run `npm install ipfs` first? SGTM.

hugomrdias:
Can you explain same use cases you targeting with change please?
Do we need to have http-client as a dep? Shouldn't we always use whatever js-ipfs depends on?

achingbrain:
> Can you explain same use cases you targeting with change please?

We use this module to test `js-IPFS` in CI, at which point we have a version we want to test it with and not the one this module is bundled with, so the time taken to install it is wasted.

The same is true for `go-IPFS` in reverse so I suppose this could be extended to the `go-ipfs-dep` dependency as well.

> Do we need to have http-client as a dep? 

It's used [here](https://github.com/ipfs/interop/blob/b7aa3131dbae6cca7112f4c0cff3cbd042a556b8/test/utils/daemon.js#L5) so yes, until that's refactored.

> Shouldn't we always use whatever js-ipfs depends on?

Preferably but `js-ipfs` doesn't export a client so the user has to keep it in sync.

PR #91, opened 2019-10-22
chore: avoid double travis build for PRs

achingbrain:
Disable building branches that aren't master - builds will happen for all PRs anyway, so this change means we don't get a branch build *and* a PR build on each PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: node_js
 cache: npm
+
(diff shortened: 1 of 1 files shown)


achingbrain:
I think the travis UI is a "build all branches" or "build no branches" thing.  This change still builds master on changes to it.

achingbrain:
image

PR #94, opened 2019-12-05
fix(ipns): fix allowOffline flag

Stebalien:


Stebalien:
@alanshaw this looks like it was a small breaking change (not really an issue, just an FYI)

alanshaw:
Aww yeah looks like that one got missed somewhere along the line :(

https://github.com/ipfs/js-ipfs-http-client/commit/fff35c811dd4eab05870d04d588c13c693975a05

PR #95, opened 2019-12-06
fix(ci): install the ipfs-http-client

Stebalien:
I'm pretty sure this _shouldn't_ be necessary but it is. The tests won't pass otherwise.


achingbrain:
Fixed by #99

Issue #92, opened 2019-10-29
CI fails because no go-ipfs doesn't accept 1024 bits anymore


daviddias:
See 
<img width="1251" alt="image" src="https://user-images.githubusercontent.com/1211152/67748623-a036e280-fa2b-11e9-986e-c4b209b28b67.png">
from https://circleci.com/gh/ipfs/go-ipfs/21905?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

This https://github.com/ipfs/interop/blob/master/test/utils/daemon.js#L12 and all the others need to change to 2048 at minimum.

Stebalien:
This has been fixed.

Issue #31, opened 2018-08-09
Should work when other IPFS nodes are running.


Stebalien:
Currently, the pubsub tests (at least) fail when my local IPFS daemon is running. Ideally, we'd use ephemeral ports for all IPFS daemons.

alanshaw:
This is likely because the tests have been coded to wait for a single message from a specific peer.

e.g. https://github.com/ipfs/interop/blob/1fd66fdce826a976882a11a52c24019e8bd10f97/test/pubsub.js#L181-L188

Issue #30, opened 2018-07-24
Debugging


Stebalien:
Please save stderr, stdout, and any errors for failed tests to files.

Stebalien:
Also, when a test fails, we need to actually *print* the error. Currently, I just get a test number and the name of the test in red with no explanation for what went wrong or why I'm even seeing this.

alanshaw:
By default mocha (the test runner) will save up the output until all the tests are done and then print the failed tests along with the error messages. If you want to fail fast and print the error immediately run `npm run test -- --bail`. Hope that helps!

Stebalien:
> If you want to fail fast and print the error immediately...

Thanks, that will definitely help.

However, it would still be nice to get the *daemon*'s stderr/stdout.

alanshaw:
In the new https://github.com/ipfs/js-ipfsd-ctl you can use the env var `DEBUG=ipfsd-ctl:daemon:std*` to get daemon logs.

PR #96, opened 2020-01-22
refactor: use new async await APIs and update ipfsd-ctl

alanshaw:
resolves https://github.com/ipfs/interop/issues/92 - does not specify `bits` anymore
resolves https://github.com/ipfs/interop/issues/31 - uses "test" profile for all daemon spawns
resolves https://github.com/ipfs/interop/issues/30 - new ipfsd-ctl exposes daemon logs via `DEBUG=ipfsd-ctl:daemon:std*`

TODO:

1. [x] fix the websocket-star tests - websocket-star was not converted to async/await and has a security vulnerability. Switch to using stardust or webrtc-star for the tests
1. [x] fix ipfsd-ctl to auto kill process if it does not stop within a timeout. The previous version of ipfsd-ctl had this feature and was killing go-ipfs daemons. I've observed here and in master that go-ipfs does not quit when running tests where it sends files to/from a js-ipfs node. @stebalien were you aware?
  * https://github.com/ipfs/js-ipfsd-ctl/pull/441
1. [ ] fix issue with ky never returning from a call to `res.json()` if the payload is BIG. IDK why but ky clones the response, which causes it to be piped to two places, the first is never consumed so for large responses the high water mark is hit and the stream stops flowing. It's similar to the problem we [worked around here](https://github.com/ipfs/js-ipfs-http-client/blob/d7eb0e8ffb15e207a8a6062e292a3b5babf35a9e/src/lib/error-handler.js#L12-L23).
  * https://github.com/ipfs/interop/pull/96/files#diff-6581b67864c5dbbaa734e512057741ceR163-R167


jacobheun:
> fix the websocket-star tests - websocket-star was not converted to async/await and has a security vulnerability. Switch to using stardust or webrtc-star for the tests

I'll work on converting this over to webrtc-star

Stebalien:
>  I've observed here and in master that go-ipfs does not quit when running tests where it sends files to/from a js-ipfs node. @Stebalien were you aware?

Go-ipfs will _try_ to wait for all in-progress commands to stop. If that doesn't work, it'll die after 10s.

We try to cancel in-progress commands but could be getting stuck. Could you file an issue somewhere? We should try killing go-ipfs with SIGQUIT sometime to see where it's getting stuck (that will dump a stack trace).

hugomrdias:
@Stebalien should we wait for the 10s or just SIGKILL please have a look at https://github.com/ipfs/js-ipfsd-ctl/pull/441#pullrequestreview-347184019

achingbrain:
Looks like it just needs the two gh urls removing from package.json and it's good to go?

@alanshaw build is green - is the fix for the `ky` issue mentioned at the top still required?

package.json
```diff
@@ -40,30 +35,32 @@
...
+    "libp2p-webrtc-star": "^0.17.3",
     "mocha": "^5.2.0",
     "multihashes": "~0.4.14",
     "p-retry": "^4.1.0",
     "pretty-bytes": "^5.1.0",
     "promisify-es6": "^1.0.3",
     "random-fs": "^1.0.3",
-    "readable-stream-buffer-stream": "~0.0.2",
-    "rimraf": "^2.6.3"
+    "readable-stream-buffer-stream": "github:alanshaw/readable-stream-buffer-stream#feat/browser-friendly",
```
achingbrain:
```suggestion
    "readable-stream-buffer-stream": "^1.0.0",
```

alanshaw:
> @alanshaw build is green - is the fix for the `ky` issue mentioned at the top still required?

The test is skipped. It still needs fixing (i.e. remove ky from ipfs-http-client) but we can probably merge this, open an issue and revisit when we've fixed the ky issue.

cc @hugomrdias

package.json
alanshaw: Pushed this update (needs package-lock.json update as well) - thanks.

hugomrdias:
> > @alanshaw build is green - is the fix for the `ky` issue mentioned at the top still required?
> 
> The test is skipped. It still needs fixing (i.e. remove ky from ipfs-http-client) but we can probably merge this, open an issue and revisit when we've fixed the ky issue.
> 
> cc @hugomrdias

yes, ill remove ky asap and come back to this issue.

achingbrain:
I'm going to merge this since this branch is basically defacto master right now.

I've opened https://github.com/ipfs/interop/issues/101 to track the skipped test.

PR #97, opened 2020-01-31
feat: pass extra cli args to npm test command

achingbrain:
Lets us keep env config in `package.json` instead of `.travis.yml` and do things like:

```console
$ npm run test:interop:browser
```

which will then invoke a npm script like:

```json
{
  "scripts": {
    "test:interop:browser": "IPFS_JS_MODULE=`pwd` npm run test:interop -- -t browser"
  }
}
```

### DIFF
--- a/bin/ipfs-interop.js
+++ b/bin/ipfs-interop.js
@@ -4,7 +4,7 @@
 
 const { spawn } = require('child_process')
 
-const proc = spawn('npm', ['test'], {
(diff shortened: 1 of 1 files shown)


achingbrain:
Simplified interop running: https://github.com/ipfs/js-ipfs/pull/2729/commits/acd2409731ae89fdd2bc30b5caec6aa774eadcbf

PR #98, opened 2020-02-04
fix: throw if no requested address is available

achingbrain:
We loop through available multiaddrs looking for certain types of address in our tests.  If there are non of the type we are looking for, we don't return anything which results in undescriptive error messages from libp2p.

The change here is to throw an error with an explicit message if, for example, no webrtc-star address is available when we expect there to be.

### DIFF
--- a/test/utils/circuit.js
+++ b/test/utils/circuit.js
@@ -67,29 +67,61 @@ exports.send = async (nodeA, nodeB) => {
   expect(buffer.slice()).to.deep.equal(data)
 }
 
-const getWsAddr = (addrs) => addrs
-  .map((a) => a.toString())
-  .find((a) => {
-    return a.includes('/ws') && !a.includes('/p2p-webrtc-star')
-  })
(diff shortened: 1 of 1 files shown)


test/utils/circuit.js
```diff
@@ -67,29 +67,61 @@ exports.send = async (nodeA, nodeB) => {
...
-const getTcpAddr = (addrs) => addrs
-  .map((a) => a.toString())
-  .find((a) => !a.includes('/ws') && !a.includes('/p2p-websocket-star'))
+const getTcpAddr = (addrs) => {
+  addrs = addrs.map((a) => a.toString())
+  const result = addrs
+    .find((a) => !a.includes('/ws') && !a.includes('/p2p-websocket-star'))
+
+  if (!result) {
+    throw new Error('No webrtcstar address found in ' + addrs)
```
alanshaw:
```suggestion
    throw new Error('No TCP address found in ' + addrs)
```

PR #99, opened 2020-02-05
chore: install ipfs, go-ipfs-dep and ipfs-http-client as dev deps

achingbrain:
This will let us run our tests without needing extra installs in .travis.yml.


alanshaw:
closes https://github.com/ipfs/interop/pull/95 ?

achingbrain:
Yes, it does 😄

PR #102, opened 2020-02-13
chore: upgrade ipfsd-ctl to the latest version

achingbrain:
Also uses one daemon factory for all node types and removes unused `find-bin.js` file and karma-webworker dep.


achingbrain:
Build is passing so I'm going to merge this to get a release out so we can remove the various dependencies on the now-merged-and-deleted async/await branch from this repo.

PR #105, opened 2020-03-01
fix: disable test with small file and raw leaves

achingbrain:
Disabled temporarily until https://github.com/ipfs/go-ipfs/issues/6940 is implemented in go-IPFS.


achingbrain:
I think we're going to go a different direction with this and still produce a non-unixfs DAG unless the user specifies metadata.

PR #108, opened 2020-03-12
fix: fix name-pubsub test

hugomrdias:
Doing a name.resolve to subscribe should not be expected to reject in the second test because both tests use the same nodes(which should be changed) and in the second test name.resolve will resolve because the nodes already now about each other.

Check this PR for CI https://github.com/ipfs/js-ipfs/pull/2932


test/ipns-pubsub.js
```diff
@@ -97,7 +97,13 @@ const subscribeToReceiveByPubsub = async (nodeA, nodeB, idA, idB) => {
   const topic = `${namespace}${base64url.encode(keys.routingKey.toBuffer())}`
 
   // try to resolve a unpublished record (will subscribe it)
-  await expect(last(nodeB.api.name.resolve(idA, { stream: false }))).to.be.rejected()
```
achingbrain: You should be able to assert on errors with something like

```javascript
await expect(last(nodeB.api.name.resolve(idA, { stream: false }))).to.eventually.be.rejected.with(/was not found in the network/)
```

hugomrdias: but that only happens in the first test not the second

achingbrain: If the order of the API calls affects how the tests run this should be made explicit in the test - an expectation to fail based on the number of invocations, or a different helper function or something like that.

achingbrain:
I think you need to update the `package-lock.json` file - CI is failing because it's not in sync with the dep updates in `package.json`.

PR #109, opened 2020-03-23
fix: all repo compatibility tests

Stebalien:
We've changed how we name keys in the keystore but the js-ipfs migration isn't quite ready. The real issue here is that the repo versions differ.

### DIFF
--- a/test/ipns.js
+++ b/test/ipns.js
@@ -79,7 +79,10 @@ describe('ipns locally using the same repo across implementations', function () 
     await publishAndResolve(goDaemon)
   })
 
-  it('should publish an ipns record to a js daemon and resolve it using a go daemon through the reuse of the same repo', async function () {
+  // FIXME: https://github.com/ipfs/js-ipfs/issues/1467
+  //
+  // Repo versions are different.
+  it.skip('should publish an ipns record to a js daemon and resolve it using a go daemon through the reuse of the same repo', async function () {
     const daemons = await Promise.all([
       daemonFactory.spawn({
         ...daemonOptions,
@@ -94,7 +97,10 @@ describe('ipns locally using the same repo across implementations', function () 
     await publishAndResolve(daemons[0], daemons[1])
   })
 
-  it('should publish an ipns record to a go daemon and resolve it using a js daemon through the reuse of the same repo', async function () {
+  // FIXME: https://github.com/ipfs/js-ipfs/issues/1467
+  //
+  // Repo versions are different.
+  it.skip('should publish an ipns record to a go daemon and resolve it using a js daemon through the reuse of the same repo', async function () {
     const daemons = await Promise.all([
       daemonFactory.spawn({
         ...daemonOptions,
--- a/test/repo.js
+++ b/test/repo.js
@@ -18,7 +18,10 @@ async function catAndCheck (api, cid, data) {
   expect(fileData.slice()).to.eql(data)
 }
 
-describe('repo', function () {
+// Repo compatibility is broken.
+//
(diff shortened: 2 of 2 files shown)


achingbrain:
Fails with:

```
  1) ipns locally using the same repo across implementations
       should publish an ipns record to a js daemon and resolve it using a go daemon through the reuse of the same repo:
     Error: Timeout of 160000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/ipfs/go-ipfs/interop/test/ipns.js)
  

  2) ipns locally using the same repo across implementations
       should publish an ipns record to a go daemon and resolve it using a js daemon through the reuse of the same repo:
     Error: Timeout of 160000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/ipfs/go-ipfs/interop/test/ipns.js)
```

@vasco-santos any ideas?

Stebalien:
Go PR: https://github.com/ipfs/go-ipfs/pull/6955

vasco-santos:
We currently use this format: https://github.com/ipfs/js-ipns/blob/master/src/index.js#L209-L219

I need to try it out and look on go changes, but cannot do it today

Stebalien:
@vasco-santos we haven't changed the way we're storing IPNS records, we've changed the way we encode _key_ names in the on-disk keystore. As far as I know, that shouldn't affect _this_ test as we're using the identity key (which isn't stored in the keystore, IIRC).

vasco-santos:
@Stebalien I understood and I think that the problem might be the key name stored in the local datastore. We store locally [here](https://github.com/ipfs/js-ipfs/blob/447b44d1b64714f5ed0cafba166ad0a4dbbb587c/packages/ipfs/src/core/ipns/publisher.js#L230)  the IPNS record, using the datastore key in [this format](https://github.com/ipfs/js-ipns/blob/master/src/index.js#L209-L219).

For what I understand, the naming should have changed and `js-ipfs` cannot get it from the datastore, in order to get the IPNS records. I need to check what is going on later to be more accurate

Stebalien:
I'm lost. We haven't changed anything about IPNS records and/or how we store/name them. We've only changed how we encode the names of private keys.

hsanjuan:
Isn't the js daemon just failing to start because of the different repo version when using same repo?

```
[23:11:33] ~/n/.bin $ ./jsipfs daemon
Initializing IPFS daemon...
js-ipfs version: 0.41.2
System version: x64/linux
Node.js version: 12.16.1
Incompatible repo version. Migration needed. Pass --migrate for automatic migration
```

Stebalien:
Oh. My bad. That explains it.

Stebalien:
I've changed this to disable _all_ repo compatibility tests as they all fail now.

achingbrain:
What feature is the repo version bump in go for?

Stebalien:
1. Fixing the bootstrap peers in the config to remove the 1024bit peers, make sure we have the new 2048bit peers, and replace all instances of /ipfs/ with /p2p/.
2. Encoding keys in the on-disk keystore using base32: https://github.com/ipfs/js-ipfs/pull/2410

hsanjuan:
should this be merged?

achingbrain:
CI is still failing, that should be resolved before merging

Stebalien:
The CI issue is unrelated. Force-pushing to rerun.

achingbrain:
The firefox circuit tests have been fixed in master by upgrading `libp2p-webrtc-star` as part of #108 - ref: https://github.com/libp2p/js-libp2p-webrtc-star/pull/212

PR #113, opened 2020-05-20
fix: go-ipfs 0.5 sync

hugomrdias:
- fix remote node ipfs http client path
- fix rabin params go-ipfs has validation and returns errors
- fix use cid instead of hash in the pin tests

### DIFF
--- a/test/files.js
+++ b/test/files.js
@@ -354,7 +354,7 @@ describe('files', function () {
 
     it('rabin chunker small chunks', () => {
       const options = {
-        chunker: 'rabin-16-16-16',
+        chunker: 'rabin-16-32-64',
         pin: false,
         preload: false
       }
--- a/.aegir.js
+++ b/.aegir.js
@@ -47,7 +47,7 @@ module.exports = {
         }, {
           type: 'go',
           test: true,
-          ipfsHttpModule: require('ipfs-http-client')
+          ipfsHttpModule: require(process.env.IPFS_JS_HTTP_MODULE || 'ipfs-http-client')
         }, {
           go: {
             ipfsBin: process.env.IPFS_GO_EXEC || require('go-ipfs-dep').path()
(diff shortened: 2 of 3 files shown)


test/files.js
```diff
@@ -354,7 +354,7 @@ describe('files', function () {
 
     it('rabin chunker small chunks', () => {
       const options = {
-        chunker: 'rabin-16-16-16',
+        chunker: 'rabin-16-32-64',
```
achingbrain, line 357: Why was this change necessary?  `rabin-16-16-16` seems to be supported by js and go:

```console
$ ipfs add --chunker=rabin-16-16-16 ./package.json 
added QmTUSudDKFYaVHLdU1wRoHYXkydQaDch4Kt8CwuVFShpiU package.json
$ jsipfs add --chunker=rabin-16-16-16 ./package.json 
added QmTUSudDKFYaVHLdU1wRoHYXkydQaDch4Kt8CwuVFShpiU package.json
```

hugomrdias: api returns validation errors, min needs to bigger than avg, avg need to be smaller than max etc

achingbrain: Which api?  The js-IPFS api?  Should probably fix that.

hugomrdias: Didn't confirm, but I'm assuming go http api we didn't change anything.
I will look into it

achingbrain: Oh, no - I think you're correct, it's go-IPFS that's changed, I think I was running an old version just now (still 0.5.x but an old version from go-ipfs master):

```console
ipfs add --chunker=rabin-16-16-16 ./package.json 
Error: incorrect format: rabin-min must be smaller than rabin-avg
```

achingbrain: @Stebalien was this change intentional?

Stebalien: This chunker config is equivalent to a fixed-sized chunker. I'm not sure if this was intentional, but I'd like to leave it as it helps detect misuse of the chunker.

hugomrdias: added this item here https://github.com/ipfs/js-ipfs/issues/3030 so we replicate in js-ipfs

PR #116, opened 2020-05-29
chore: add gossipsub grafting delay

achingbrain:
Because of https://github.com/libp2p/go-libp2p-pubsub/issues/331 we have to wait for a the gossipsub peer to become grafted before sending messages we expect to be received.  There's no way to tell from the API which peers are grafted so we just have to add an arbitrary wait and hope it happens during that time window.

Also refactors the pubsub tests to remove duplicated code and the ipns-pubsub tests to use named variables instead of array indices for readability.


achingbrain:
This doesn't seem to fix the problem after all.

PR #136, opened 2020-07-16
Test imported js-ipfs ipns key pubsub'ing to, and resolving on go-ipfs

DougAnderson444:
## Context

After having issues publishing to `ipns` using an imported `key` in `js-ipfs`, then (with pubsub enabled) attempting to resolve that `ipns key` on `go-ipfs`, @vasco-santos [proposed](https://discuss.ipfs.io/t/cant-access-published-ipns-record-config-problem/5476/17) submitting a PR with failing tests. This is the PR to show where it's failing.

### Prior Art

This test is based off [https://github.com/ipfs/interop/blob/master/test/ipns-pubsub.js#L89-L115](https://github.com/ipfs/interop/blob/master/test/ipns-pubsub.js#L89-L115)

### Referenced People and Conversations:

https://discuss.ipfs.io/t/cant-access-published-ipns-record-config-problem/5476/17
https://discuss.ipfs.io/t/ipns-not-resolvable-from-other-node/8624/3
https://github.com/ipfs/go-ipfs/issues/6360  

I've been trying to figure this out with @aschmahmann who has been doing great work on the go-side of things, and his upcoming changes may fix this issue. For now, here's the failing test on it.

### Failing tests

The test fail:
1. When we try to ipfsHttpApi.name.resolve({ key: importedKey }). go-ipfs API doesn't resolve it
2. The pubsub /record/ and /ipns/ records don't get created on the resolving go-node when an imported js key is used? That's weird.
3. Lastly of course, after the ipfsBrowser.name.publish() completes, the records are not pubsub'd over nor does anything resolve (likely due to the first two failing steps).

When the imported key is replaced with 'self' key, everything works fine, as expected.

#### Other note

I added `"test:node-ipns"` just so I could run this test by itself, left it there for anyone who needs the same functionality while reviewing this test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "lint": "aegir lint",
     "test": "cross-env IPFS_REUSEPORT=false aegir test",
     "test:node": "cross-env IPFS_REUSEPORT=false aegir test -t node -f test/node.js",
+    "test:node-ipns": "cross-env IPFS_REUSEPORT=false aegir test -t node -f test/node-ipns.js",
     "test:browser": "cross-env IPFS_REUSEPORT=false aegir test -t browser --no-cors -f test/browser.js"
   },
   "repository": {
--- a/test/js-publish-go-resolve-pubsub.js
+++ b/test/js-publish-go-resolve-pubsub.js
@@ -1,0 +1,288 @@
+/* eslint-env mocha */
+"use strict";
+
+const { fromB58String } = require("multihashes");
+const base64url = require("base64url");
+const ipns = require("ipns");
+const delay = require("delay");
+const last = require("it-last");
+const pRetry = require("p-retry");
+const waitFor = require("./utils/wait-for");
+const { expect } = require("./utils/chai");
+const daemonFactory = require("./utils/daemon-factory");
+
+const retryOptions = {
+  retries: 5,
+};
+
+const namespace = "/record/";
+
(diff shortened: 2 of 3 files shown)


aschmahmann:
@DougAnderson444 do you mind testing against this branch of go-ipfs to see what happens https://github.com/ipfs/go-ipfs/pull/7549? If you don't have time or are having trouble lmk and I'll get to it when I get a chance.

DougAnderson444:
I didn't get very far -- tried to `npm install` using `"go-ipfs": "git+https://github.com/ipfs/go-ipfs#feat/ipns-stop-pk-resolve"` but `npm` sure didn't like that.

jacobheun:
@aschmahmann I ran this against your branch and things are passing. I did have to make some local fixes to the tests. `key` is not defined in one of them.

The tests should get cleaned up before this gets merged. Some of the later tests are dependent on the earlier ones which makes this suite fragile to future updates. It also makes the tests harder to follow.

_fyi the easiest way to run against a go branch is to build it locally and use the env variable for the path `IPFS_GO_EXEC=$(which ipfs) npx aegir test -t node -f test/node-ipns.js`_

Also, I found a bug in the js http client. Streaming can't be overridden (it's always true) which is giving us some false positives. I'll submit a fix for that.

aschmahmann:
>  I ran this against your branch and things are passing

@jacobheun great, so I'll merge that branch then.

PR #141, opened 2020-07-22
fix tests for new IPNS name formatting

petar:


petar:
@jacobheun I need a little help on this PR. I've made a small change, which seems unrelated to why the PR is not building in CI. If you click through, you'll see it is not building on Mac and Windows, but I can't understand the errors. They seem to be coming from the guts of js-ipfs. Advice appreciated, thanks!

jacobheun:
The mac invalid issue is a known problem that results in the tests being flakey. We haven't figured out the root cause of that but we typically just rerun the build for that failed test when it happens.

I'm not sure why windows is hitting a teardown issue trying to access the go repo. This looks like a recent failure that also occurred on the latest master build. @achingbrain have you seen the EPERM issue before on windows?
```
Error: EPERM: operation not permitted, lstat 'C:\Users\travis\AppData\Local\Temp\go_ipfs_UhteN95f6K1jZYFzfBEyx\keystore'
```

PR #144, opened 2020-07-29
test: reduce ipns pubsub test duration

jacobheun:
The `ipns-pubsub` tests were taking 2+ minutes to complete. There was a 60 second delay in the before test. I assume this was added previously due to some timing issues, but this isn't necessary. Additionally, I'm not sure why the ipns resolve statements were being done serially, this just adds to the test time (1 test regularly takes about 60s to complete, likely due to the republish interval but I'm not certain). 

This update removes the unnecessary delay in the `before` and lets the pubsub and regular resolve happen in parallel which **reduces the time by ~2mins**.

| **Before** | **After** |
| ------------- | ------------- |
| image | image |

### DIFF
--- a/test/ipns-pubsub.js
+++ b/test/ipns-pubsub.js
@@ -111,7 +111,7 @@ const subscribeToReceiveByPubsub = async (nodeA, nodeB, idA, idB) => {
   await waitFor(() => subscribed === true, (50 * 1000))
   const res2 = await last(nodeB.api.name.resolve(idA))
 
-  expect(PeerID.createFromCID(res1.name).toString()).to.equal(PeerID.createFromB58String(idA).toString()) // Published to Node A ID
+  expect(PeerID.createFromCID(res1.name).toString()).to.equal(PeerID.createFromCID(idA).toString()) // Published to Node A ID
   expect(res2).to.equal(ipfsRef)
 }
 
--- a/test/ipns-pubsub.js
+++ b/test/ipns-pubsub.js
@@ -27,8 +27,8 @@ describe('ipns-pubsub', function () {
   let nodes = []
 
   // Spawn daemons
-  before(async function () {
-    this.timeout(10e3)
+  before('create the nodes', async function () {
+    this.timeout(20e3)
     nodes = await Promise.all([
       daemonFactory.spawn({
         type: 'go',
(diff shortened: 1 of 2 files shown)


test/ipns-pubsub.js
```diff
@@ -53,14 +52,12 @@ describe('ipns-pubsub', function () {
...
-    // Remove this when js-ipfs has a DHT
-    await nodes[0].api.swarm.connect(nodes[2].api.peerId.addresses[0])
-
-    console.log('wait for republish as we can receive the republish message first') // eslint-disable-line
-    await delay(60000)
+    // Remove the second connect when js-ipfs runs a DHT server
+    await Promise.all([
+      nodes[0].api.swarm.connect(nodes[1].api.peerId.addresses[0]),
+      nodes[0].api.swarm.connect(nodes[2].api.peerId.addresses[0])
+    ])
```
vasco-santos, line 60: In the IPNS over pubsub flow, there is a `initialBroadcastInterval` of 60 seconds. If we don't wait for it, the tests below might be impacted by the initial broadcast and not by the publish that happened in the test.
I think the solution is to allow this `initialBroadcastInterval` to be configured, both in go and js

vasco-santos: Actually, It can be configured in JS: https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs/src/core/ipns/republisher.js#L72-L78

jacobheun: Does the initial broadcast **only** happen after 60 seconds? The tests complete in < 5 seconds.

vasco-santos: > Does the initial broadcast only happen after 60 seconds?

By default yes. I remember it could happen to have false positives when this got in. But maybe, that's when we were also testing with the DHT

jacobheun: Maybe this was due to the tests originally taking a very long time? Since the whole suite now finishes in under 5 seconds the first republish shouldn't be doing anything. I should probably remove the timeouts in these tests. If they do ever start taking 1 minute to resolve this could potentially create false positives.

vasco-santos: That sounds good!

jacobheun: I removed the larger timeouts and added smaller ones for the startup and initial connects. With the updated timeouts the whole suite (aside from teardown) has a max allowed time of 45 seconds (30 for boot and 5*3 for tests).

PR #203, opened 2020-12-14
chore: update websockets to 0.15

vasco-santos:
This PR updates websockets usage to `0.15` per [libp2p migration v0.29-v0.30.md#development-and-testing](https://github.com/libp2p/js-libp2p/blob/0.30.x/doc/migrations/v0.29-v0.30.md#development-and-testing).

Local websocket dial from a browser should not be filtered out for testing purposes


achingbrain (approved):
Changes LGTM.

1. Do we need libp2p@0.30.0 in js-ipfs before this can be released?
2. What happened to the package-lock file?  670 additions and 27k deletions?

vasco-santos:
> Do we need libp2p@0.30.0 in js-ipfs before this can be released?

No, it does not need it! libp2p already supported passing the transport configuration. We just got this together with `0.30` to notice the potential breaking usage if examples/tests use local addresses. Libp2p integration PR in IPFS needs this as it uses the new websockets as well.

> What happened to the package-lock file? 670 additions and 27k deletions?

I retried doing it, but the same result. I just installed `libp2p-websockets` and it auto updated like that.
I think that before it relied on libp2p-websockets via other modules (like IPFS) that have it has dependencies. But now, as we install it in this repo it changed the JSON structure to have the websockets in the top level.

achingbrain:
> I retried doing it, but the same result

What version of node are you running?

achingbrain:
Shipped in `ipfs-interop@4.0.1` btw

vasco-santos:
I am with `14.15.2`

PR #223, opened 2021-02-06
feat: new aegir

hugomrdias:


test/utils/daemon-factory.js
```diff
@@ -2,22 +2,20 @@
...
-  ipfsModule = require('ipfs')
-}
 
-if (process.env.IPFS_JS_HTTP_MODULE) {
+let ipfsHttpModule
+let ipfsModule
+try {
   ipfsHttpModule = require(process.env.IPFS_JS_HTTP_MODULE)
-} else {
+} catch (err) {
```
achingbrain, line 10: We should be able to use [optional catch binding](https://2ality.com/2017/08/optional-catch-binding.html) now?

```suggestion
} catch {
```

test/utils/daemon-factory.js
```diff
@@ -2,22 +2,20 @@
...
   ipfsHttpModule = require('ipfs-http-client')
 }
 
+try {
+  ipfsModule = require(process.env.IPFS_JS_MODULE)
+} catch (err) {
```
achingbrain, line 16: Same as the other try/catch

```suggestion
} catch {
```

package.json
```diff
@@ -63,6 +64,12 @@
     "rimraf": "^3.0.0",
     "uint8arrays": "^2.0.5"
   },
+  "eslintConfig": {
+    "extends": "ipfs",
+    "ignorePatterns": [
+      "!.aegir.js"
```
achingbrain, line 70: Why are we not linting this file?

achingbrain (changes requested):
LGTM, a few small nits

package.json
hugomrdias: this actually tell the linter to **lint** the file, double negative lol

hugomrdias: eslint ignores dot files by default we need to tell the linter to not ignore .aegir.js inside the ignore config 🤷🏻‍♂️

PR #362, opened 2021-09-14
chore: update ipns

achingbrain:


achingbrain:
Fixed in https://github.com/ipfs/interop/pull/367

PR #376, opened 2021-10-01
feat: expand dht tests

achingbrain:
Adds additional tests from #36 as well as some extras requested in the PR comments.

Refs: https://github.com/ipfs/js-ipfs/issues/3905


lidel:
Continued in https://github.com/ipfs/interop/pull/410

PR #382, opened 2021-10-13
fix: update tests to support go-ipfs automatic sharding

aschmahmann:


test/files.js
```diff
@@ -9,7 +9,10 @@ import all from 'it-all'
 import last from 'it-last'
 import { expect } from 'aegir/utils/chai.js'
 
-const SHARD_THRESHOLD = 1000
+const SHARD_THRESHOLD = 2000
+// Use this to help push the shard threshold over the go-ipfs defaults
+// without just adding lots of files
+const shardFilePadding = "".padStart(100, '0')
```
aschmahmann, line 15: @achingbrain we could figure out adding a configurable sharding threshold in go-ipfs. However, this seemed like it wasn't super painful and wouldn't make tests take too much longer. WDYT?

package.json
```diff
@@ -97,7 +97,7 @@
   ],
   "devDependencies": {
     "go-ipfs": "^0.9.1",
-    "ipfs": "^0.59.0",
+    "ipfs": "^0.60.0",
```
aschmahmann, line 100: @achingbrain how do I test the fix from  ipfs/js-ipfs#3921?

achingbrain: This test has just the right number of magic file names to trigger the hashing behaviour that caused the bug fixed in that PR: https://github.com/ipfs/js-ipfs/pull/3936/files#diff-77ef56bdffd9bf9f144b53c8ca4937504b82421c88ddf508ed0d11f4d25904f3R908-R949

achingbrain: Though is that the sort of thing that you would test here?  It doesn't seem like an interop issue, just a bug in js-ipfs.

test/files.js
achingbrain: Having a configurable sharding threshold makes things like the test added as part of https://github.com/ipfs/js-ipfs/pull/3936 more feasible.  One or two tests like this are not super painful, but twenty or thirty would be.

achingbrain: It also means you can do the whole test in code instead of having to deal with importing car files to get HAMTs that are \*just so\*.  Doing it all in code means it's easy to run in environments without filesystems like browsers.

lidel:
Continued  with approach from #393  in https://github.com/ipfs/interop/pull/400

PR #393, opened 2021-11-10
fix: use new go-ipfs sharding option

aschmahmann:
This will lower the threshold to the minimum for compatibility with existing tests as an alternative to #382 while https://github.com/ipfs/js-ipfs/issues/3921 is being investigated


lidel:
Continued in https://github.com/ipfs/interop/pull/400

PR #397, opened 2021-11-17
feat: scripts/custom-runtime.sh

lidel:
This PR extracts the glue code from #387  that simplifies running interop against unreleased code of go-ipfs and js-ipfs packages from lerna monorepo.

Goal here is to make it easier to land changes by demonstrating green CI and/or speed up local dev.

## DEMO

See  it in action in #387 where it takes care of running against specific git revisions of go-ipfs, js-ipfs(-http-server) and js-ipfs-http-client :robot: 

(next) I am going to use this for unblocking libp2p as well.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,13 +4,25 @@ on:
     branches:
       - master
 
+# To run CI against unrelased go-ipfs or js-ipfs-* code (eg. wip PR),
+# uncomment env below and define git revisions in ./scripts/custom-runtime.sh
+
+#env:
+  #IPFS_GO_EXEC: /tmp/go-ipfs/cmd/ipfs/ipfs
+  #IPFS_JS_EXEC: /tmp/js-ipfs/packages/ipfs/src/cli.js
+  #IPFS_JS_MODULE: /tmp/js-ipfs/packages/ipfs/dist/cjs/src/index.js
+  #IPFS_JS_HTTP_MODULE: /tmp/js-ipfs/packages/ipfs-http-client/dist/cjs/src/index.js
+
 jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.16'
+      - uses: actions/setup-node@v2
         with:
           node-version: 16
       - uses: actions/cache@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -4,26 +4,39 @@ on:
     branches:
       - master
 
+env:
+  # To run CI against unrelased go-ipfs or js-ipfs-* code (eg. wip PR),
+  # uncomment below and define git revisions in ./scripts/custom-runtime.sh
+  #IPFS_GO_EXEC: /tmp/go-ipfs/cmd/ipfs/ipfs
+  #IPFS_JS_EXEC: /tmp/js-ipfs/packages/ipfs/src/cli.js
+  #IPFS_JS_MODULE: /tmp/js-ipfs/packages/ipfs/dist/cjs/src/index.js
+  #IPFS_JS_HTTP_MODULE: /tmp/js-ipfs/packages/ipfs-http-client/dist/cjs/src/index.js
```
aschmahmann, line 13: This works and is documented ✔️

What's the advantage of doing this vs always testing against custom-runtime.sh, but sometimes checking out the release tag/branch and other times checking out a Git hash? We could pass two versions into the script (e.g. `custom-runtime.sh v0.10.0 v0.60.0`) and then have the script export environment variables.

Perhaps this is a pain and could be circumventing by passing the variables into the interop test. If so this seems like a fine change to make at a later date.

aschmahmann (changes requested):

README.md
```diff
@@ -24,12 +24,6 @@ $ npm install -g ipfs-interop
 $ ipfs-interop
 ```
 
-### Run a particular test locally
-
-```console
-$ node bin/ipfs-interop.js -- -t node --grep {substring-test-name}
-```
-
```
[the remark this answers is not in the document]
lidel: messed rebase, should be fixed now

aschmahmann (approved):
I still have questions for if/how we could do this differently. However, this does the job, thank you 🙇

.github/workflows/test.yml
lidel: The problem with `custom-runtime.sh` is that (as it is right now) builds dependencies from scratch, which takes  ~6min.
This would slow down CI in go-ipfs and js-ipfs repos where ipfs/interop also runs. 

We can make it smarter, but for now i just wanted to ship mvp so the next person has head-start.

aschmahmann: SGTM, thanks. I forgot how long the js-ipfs builds can take.

PR #398, opened 2021-11-17
feat: go-libp2p v0.16.0

### LINKED ISSUE #396: update interop tests for go-libp2p v0.16
Due to the introduction of circuit v2, (for now) the node behind the firewall has to be configured to use the relay node (as a static relays).
The configuration change needed is pretty straightforward: https://github.com/ipfs/go-ipfs/blob/fde7d01a790254ae23e666dcf2eb9ec16a53fe7c/test/sharness/t0182-circuit-relay.sh#L27-L36.

However, I'm not sure how to do the change here. I assume we'd have to create a `createGoRelay` function: https://github.com/ipfs/interop/blob/b361f74924892cc9018241f1a502127100e5a579/test/utils/circuit.js#L57-L72
and then change the `createGo` function such that it uses that relay.

lidel:
This PR aims to fix the interop tests in https://github.com/ipfs/go-ipfs/pull/8522 and close  #396

- [x] switch to go-ipfs from https://github.com/ipfs/go-ipfs/pull/8522
- [x] refactor`test/utils/circuit.js` to use new go-ipfs-config keys
- [x] disable JS-GO circuit combinations that are not possible due to missing [circuit v2](https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md) support in JS
- [x] make `node bin/ipfs-interop.js -- -t node --grep "circuit"` pass


lidel:
Refactored GO side of things to leverage [circuit v2](https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md) and  ensure correct relay is used by setting `StaticRelays` (both added in https://github.com/ipfs/go-ipfs/pull/8522).

Unfortunately JS side of things (js-ipfs running in nodejs or in browser) does not talk circuit v2  and some combinations fail to pass. Landing this in time for go-ipfs 0.11 is unlikely and I think we need to `skip` them for now – marked all cases with `FIXME` keyword so we don't forget + will fill issue in js-ipfs repo after this PR is reviewed.

---

For future reference, combinations which currently do not pass:

- terse notation:
  `a-r-b`  – node `a` talking to node  `b` over relay `r` 
  `js` is js-ipfs in node
  `browser` is js in a browser context (a regular webpage, for sake of making this list short)
  `go` is go-ipfs with circuit v2 (acting as a relay client or providing the service)
  💚 cigar (circuit works, both connect and send)
  💢 no cigar (failure, either connect or send)

- `test/circuit/all.js`
  - 💚 go-go-go
  - 💢 js-go-go
  - 💢 go-go-js
  - 💢 js-go-js
  - 💚 go-js-go
  - 💚 js-js-go
  - 💚 go-js-js
  - 💚 js-js-js

- `test/circuit/browser.js`
  - 💢 browser-go-js
  - 💢 browser-go-go
  - 💚 browser-js-js
  - 💚 browser-js-go
  - 💢 js-go-browser
  - 💢 go-go-browser
  - 💚 js-js-browser
  - 💚 go-js-browser
  - 💢 go-browser-browser
  - 💚 js-browser-browser 
  - 💚 browser-browser-go 
  - 💚 browser-browser-js

lidel:
Sidenote:
- CI failure is due to unrelated sharding changes in go-ipfs 0.11
- When we get  #393  to pass, this PR should be rebased on top of that.

marten-seemann:
There are 2 reasons why a test might fail here:
1. JS doesn't speak circuit v2.
2. If a Go node is connected to another peer via a v2 relay, it will only _open_ streams for certain protocols (most notably, the hole punching protocol). That means that although a connection was successfully established, we might not be able to use it to transfer a file.

We probably should have another set of tests using [relayd](https://github.com/libp2p/go-libp2p-relay-daemon) as a relay, configured as a v1 relay. With that configuration, all tests that we currently have should pass.

lidel:
I've split circuit tests into v1 and v2 sections. v1 tests `connect` and `send`, v2 only tests `connect`. 

Tests of circuit  v1 and v2  work, but there are annoying issues which block CI from being green:
- spawning relayd fails in electron because of some ENS/CJS nonsense (needs investigation)
- there is an unrelated failure in `hamt shards`  test - most likely go-ipfs from update-libp2p pulled in some changes

@achingbrain  mind eyeballing this PR? 

Note:
- kept old tests  + added relayd ones in  node
- we don't have JS orchestration for relayd in browsers (go-ipfs/js-ipfs looks easy, because they have a mature setup based on aegir/playwright/ipfsd-ctl where js/go-ipfs is spawned remotely)
- some tests had to be disabled because go-ipfs only support v2

lidel:
CI was failing due to sharding changes, we continue this in https://github.com/ipfs/interop/pull/400 (CI is finally green there)

PR #400, opened 2021-11-23
feat: go-libp2p 0.16 and autosharding

### LINKED ISSUE #396: update interop tests for go-libp2p v0.16

aschmahmann:
This is just #398 and #393 combined [and using joint branch from https://github.com/ipfs/go-ipfs/pull/8563] because the bubbling of go-datastore through go-ipfs effected both the libp2p and sharding changes such that they're needed together in order to pass

Closes #396


lidel (approved):
Fixed relayd on CI, and switched to go-ipfs from https://github.com/ipfs/go-ipfs/pull/8563, all green.

PR #407, opened 2021-12-01
chore: go-ipfs 0.11.0-rc1 + libp2p-relay-daemon 0.1.0

lidel:
This is a sandbox to figure out next steps for #387  on the js-ipfs side

- [x] switch to  go-ipfs 0.11.0-rc1
- [x] switch to  libp2p-relay-daemon 0.1.0
- [ ] ci is green because we still use unreleased js-ipfs with pubsub changes


lidel:
Continued in https://github.com/ipfs/interop/pull/387 as we need pubsub changes from there.

PR #462, opened 2022-05-09
feat!: upgrade to esm libp2p/ipfs

achingbrain:
Updates deps and adds types to tests.

BREAKING CHANGE: needs ESM release of ipfs/libp2p

PR #493, opened 2022-09-07
chore: update to ipfsd-ctl@12.0.2

SgtPooki:


lidel (changes requested):
@SgtPooki   ipfsd-ctl@12.0.2 seems to cause tests to fail in Electron runtimes (both main and renderer): 
https://github.com/ipfs/interop/runs/8236024242?check_suite_focus=true#step:5:232

achingbrain:
This will be fixed by #462 which is almost there, just need to figure out why the go-go relay tests are timing out in CI.  Obviously everything passes locally 😢

SgtPooki:
@lidel @achingbrain The tests are reproducible locally with [`npm run test -- -t electron-main --bail -f dist/cjs/node-test/*js`](https://github.com/ipfs/interop/actions/runs/3009741011/workflow#L133), but updating it to `npm run test -- -t electron-main --bail -f dist/esm/node-test/*js` succeeds locally.

Do we still need to be testing against cjs?

SgtPooki:
Hopefully unblocked by wrapping ipfsd-ctl with a local package at test/packages/ipfsd-ctl-wrapper . At least the tests succeed locally(tested `npm run test -- -t electron-main --bail -f 'dist/cjs/node-test/*js' --debug -- --full-trace` and `npm run test -- -t electron-main --bail -f 'dist/esm/node-test/*js' --debug -- --full-trace`), hopefully CI passes. Also, I updated CI to run tests against both CJS and esm

SgtPooki:
the electron-main bypassed the error. https://github.com/ipfs/interop/runs/8257051079?check_suite_focus=true. We should be good to go.

achingbrain:
> Do we still need to be testing against cjs?

No, we should be ESM only, no more dual publishing.  Only forwards!

